Issue #2, opened 2026-09-02
XLR Wave Mk2 lsusb


BenjyEX3:
**As you requested!**
**Also curious on if there is a way to have the software NOT deal with ANY of the audio output, I just want to use it for my xlr mic and use my fosi ZH3 for the audio, Thanks.**

Bus 005 Device 004: ID 0fd9:00b6 Elgato Systems GmbH Elgato Wave XLR MK.2
Negotiated speed: High Speed (480Mbps)
Device Descriptor:
bLength                18
bDescriptorType         1
bcdUSB               2.00
bDeviceClass            0 [unknown]
bDeviceSubClass         0 [unknown]
bDeviceProtocol         0
bMaxPacketSize0        64
idVendor           0x0fd9 Elgato Systems GmbH
idProduct          0x00b6 Elgato Wave XLR MK.2
bcdDevice            2.10
iManufacturer           1 Elgato
iProduct                2 Elgato Wave XLR MK.2
iSerial                 3 AAY4I55111P6X2
bNumConfigurations      1
Configuration Descriptor:
bLength                 9
bDescriptorType         2
wTotalLength       0x019e
bNumInterfaces          5
bConfigurationValue     1
iConfiguration          0
bmAttributes         0x80
(Bus Powered)
MaxPower              500mA
Interface Association:
bLength                 8
bDescriptorType        11
bFirstInterface         0
bInterfaceCount         3
bFunctionClass          1 Audio
bFunctionSubClass       0 [unknown]
bFunctionProtocol      32
iFunction              16 Elgato Wave XLR MK.2
Interface Descriptor:
bLength                 9
bDescriptorType         4
bInterfaceNumber        0
bAlternateSetting       0
bNumEndpoints           1
bInterfaceClass         1 Audio
bInterfaceSubClass      1 Control Device
bInterfaceProtocol     32
iInterface             16 Elgato Wave XLR MK.2
AudioControl Interface Descriptor:
bLength                 9
bDescriptorType        36
bDescriptorSubtype      1 (HEADER)
bcdADC               2.00
bCategory               8
wTotalLength       0x0073
bmControls           0x01
Latency control Control (read-only)
AudioControl Interface Descriptor:
bLength                 8
bDescriptorType        36
bDescriptorSubtype     10 (CLOCK_SOURCE)
bClockID                7
bmAttributes            3 Internal programmable clock
bmControls           0x05
Clock Frequency Control (read-only)
Clock Validity Control (read-only)
[...]

PR #4, opened 2026-09-02
Harden routing, device control, and deck integration

CarinaSchoppe:
I came across OpenXLR on Reddit (cuz of the post you made on the Elgato subreddit) and pretty much immediately thought: finally, I can properly use my Stream Deck + XLR setup on Linux. So I pulled it down and started tinkering with it here and there.

On my current Arch-based setup the latest build did not behave quite as reliably as I wanted, and I ran into a few rough edges / buggy cases. Rather than jumping straight into larger feature work, I spent some time hardening the existing paths first and laying a better foundation for further work (if you don't mind ofc 🙂).

Now to what I did

This PR mainly adds reliability, safety, test coverage and some additional fail-safe behavior around the existing architecture:

added a GitHub Actions CI pipeline for restore/build/test plus validation of the Stream Deck/OpenDeck plugin and packaged service files
added an xUnit test project covering device capability checks, PipeWire stereo-pair routing, legacy profile compatibility and diagnostics redaction
hardened the Stream Deck/OpenDeck daemon connection with guarded reconnects, exponential backoff, stale-socket protection and safer command sending
made deck controls capability-aware so unsupported hardware controls are not presented or treated as if they succeeded
added server-side device capability validation, including second-XLR / second-headphone checks, and made profile recall respect unsupported controls and gain lock
tightened USB control transfers so short reads/writes are treated as errors instead of silently accepting partial data
hardened external process execution (amixer, PipeWire helpers, etc.) with stderr capture, timeouts and process termination instead of potentially hanging indefinitely
fixed a few PipeWire routing edge cases, especially stereo-pair selection, Aux routing and hardware-output exposure, so missing port pairs no longer fall back to the wrong outputs
clarified monitor-output/profile semantics so an explicit empty output list really means “disconnect all”, while older profiles remain backwards-compatible
reworked WebSocket broadcasting around a bounded per-client send queue, message size/type validation and send timeouts so slow/stuck clients cannot build up unbounded work or memory
improved diagnostics privacy by using restricted temporary directories, adding a privacy notice, and redacting common user/host/process/device-serial information before packaging diagnostics
added some systemd sandboxing/hardening (NoNewPrivileges, ProtectSystem, kernel/control-group protections, etc.)
cleaned up a couple of Linux/UI compatibility details along the way
Additional DSP / PipeWire hardening

I also ran into a particularly nasty failure mode around the host-side DSP chain: if an optional plugin was missing or a PipeWire filter failed to come up correctly, rebuilding the graph could potentially remove the existing working microphone route before discovering that the replacement could not be created.

So I reworked that path to fail much more safely:

input/filter graph changes are now effectively transactional: the replacement graph is built and validated first while the existing audio path remains active
the old microphone route is only removed after the new route, filter nodes and required PipeWire links have successfully come up
if creating the replacement graph fails, newly created nodes/links are rolled back and the previous working route stays intact
healthy direct input feeds can be reused instead of unnecessarily recreating identical PipeWire links
filter-chain rebuilding now uses unique generations/node names to avoid collisions while swapping graphs
route self-healing was generalized from just the software low-cut path to all relevant filter chains, including inserts and software ClipGuard
filter holder processes are now checked while waiting for PipeWire ports, so a process that crashes during startup is detected immediately instead of eventually returning a half-created filter
redirected stdout/stderr from long-lived pw-cli filter holders is continuously drained so a filled pipe cannot freeze the process or block the mixer
failed filter creation now cleans up the half-created process/module and returns the actual stderr/stdout information where possible
ClipGuard specifically

I also tightened up software ClipGuard quite a bit.

The XLR Dock's software ClipGuard is a post-ADC hard limiter, so I corrected the wording/documentation around that as well: it can protect the downstream PipeWire/mixer path from overload, but obviously cannot repair clipping that already happened in the analog preamp or ADC. Mic gain still needs some headroom.

More importantly, ClipGuard now fails safely when its optional LADSPA dependency is unavailable:

OpenXLR checks for the required hard_limiter_1413 LADSPA plugin before touching the live graph
it checks LADSPA_PATH, the user's LADSPA directory, standard /usr/lib / /usr/local locations and Debian-style multiarch LADSPA locations
if swh-plugins is missing, enabling ClipGuard is rejected with an actionable error instead of breaking/replacing the current microphone route
saved settings/scenes requesting ClipGuard no longer prevent the mixer from starting when the optional plugin is missing
the daemon now exposes ClipGuard availability and its dependency/load error as part of mixer state
the UI disables the ClipGuard control when the feature is unavailable and shows the actual reason in the tooltip
the Stream Deck/OpenDeck plugin treats unavailable ClipGuard as an unsupported/disabled target instead of optimistically displaying it as OFF
after a mixer command is rejected, the WebSocket client is sent the authoritative state again so optimistic UI/deck controls immediately snap back to reality instead of appearing enabled when the command actually failed
added dedicated tests for LADSPA plugin discovery and the missing-ClipGuard dependency/failure path
Documentation / protocol cleanup

While going through this I also cleaned up some of the research/protocol documentation so old investigation notes don't accidentally look like current implementation truth:

clearly marked the original research document and capture plan as historical investigation material
pointed current implementation documentation at the authoritative second Wave XLR Pro capture/control map
corrected some superseded first-capture assumptions and removed outdated/provisional control labels
documented the current Wave XLR Pro implementation, including both XLR inputs, both headphone outputs, USB Aux, output routing and the commit sequence
clarified that the XLR Dock has no onboard voice-processing DSP rather than implying that absolutely no state/control exists in firmware
updated installation notes to explicitly mention swh-plugins for software ClipGuard and lsp-plugins-lv2 for LV2 inserts
TL;DR

This is mostly groundwork rather than an attempt to redesign the project.

My goal was to make the current implementation a bit more predictable on a rolling-release Arch environment, make failures explicit instead of silent, and get tests/CI in place so it is safer to keep building on top of it.

The later changes especially focus on one principle: an optional DSP feature or plugin failing should never cost you a currently working microphone route.

So instead of tearing down the current graph and hoping the replacement works, OpenXLR now tries to build and validate the new path first and only swaps over once it knows the replacement is actually usable.

On the side

Really cool project — it was exactly the kind of thing I had been hoping existed for this hardware, so I ended up diving into the code pretty much straight away. 🙂

Cheers,
Carina


emaspa:
You merged main yourself while I was preparing the same thing, so I built on your merge instead. Pushed three commits to this branch (maintainer edits):

**c725d7e** fixes a regression from the transactional input wiring: `WireInputFeedsLocked` threw whenever `RouteInputToChannel` returned no pairs. The default config always defines XLR 1, XLR 2 and Aux In (pairs 0, 1, 2) and nothing filters channels by device, so every stereo interface (XLR Dock, Wave XLR, MK.2) failed the whole build at XLR 2 and lost the submixer; main left those channels unwired. They stay unwired now, and the throw remains for a filter chain whose own ports cannot be linked. The same commit restores the short-read tolerance in the MK.2 device class only: that family (Wave XLR MK.2, and the XLR Dock MK.2 since 0.1.10) has not been run on hardware and the block lengths come from a capture, so a strict length check would turn an unknown into a poll failure every second. The strict check stays on the verified devices.

**b5b0463** drops `ProtectKernelModules=` from the unit, the NixOS module and the UI-written unit. In a user service it fails at the CAPABILITIES step ("Failed to drop capabilities: Operation not permitted"), so the daemon would not start at all. The other six options start fine and leave ~/.config writable (tested with the real daemon under a user manager on Ubuntu 26.04, which restricts unprivileged user namespaces).

**4ccd13b** carries three of your protocol-doc corrections that the merge lost: the verified mic-bit map in the implementation note, the commit-block label for block 0x0003, and compressor2 in place of the removed polarity2.

Left for you, in decreasing order of impact:

1. `WebSocketHub`: when the 32-slot send queue is full the send returns a faulted task with `InvalidOperationException`, but `HandleAsync` only catches `WebSocketException`, so a client stalled for about two seconds (meters push at 15 Hz) that then sends a command takes the exception out of the pipeline. The pump also exits on a send failure without completing the channel writer, so a send that passed the state check just before `Abort()` never completes.
2. Profile recall with the gain lock set returns the informational "gain remained unchanged" string through the error slot, so the client gets an `error` message and no success broadcast even though the profile loaded.
3. Diagnostics redaction is substring-based: a three-letter username that occurs in the graph dump (`clock.max-quantum` for "max") corrupts it. And `device-blocks.json` is written unredacted while PRIVACY.txt says serials are redacted.
4. The `_pw.UnlinkNodes(input, ch.SinkName)` call before inserting a chain was dropped; a direct link not tracked in `_inputFeeds` would now survive alongside the chain and double the unfiltered signal.

Everything else checked out: the transactional rebuild, the process timeouts and pipe draining, the stereo-pair selection for Aux, the plugin reconnect logic, the tests and the CI workflow. Thanks for the work on this.

CarinaSchoppe:
Haha yeah saw that you tried doin something or at least I had the gut feeling.

guess I was just quicker, even though you"re pumping commits out like crazy, thought I might help a little.

CarinaSchoppe:
Btw.
I'll fix the issues prob. tomorrow.
Thx for checking in on my pull req. to quickly.

emaspa:
One more commit on the redaction, prompted by the first MK.2 diagnostics on #2: PipeWire embeds the USB serial in node and card names (`alsa_input.usb-Elgato_..._<serial>-00...`), so it was still in the graph dump, the sink and source listings and mixer.json after the `device.serial` field redaction. The serials of all attached USB devices are now read from `/sys/bus/usb/devices/*/serial` and replaced wherever they occur. The bare user-name replacement is gone for the reason in my earlier point 3; the home path covers where it actually shows up. Two tests added.

emaspa:
Merged. Since main was moving fast today (three releases) I did the four remaining items myself rather than have you chase a moving branch, in b09ce38: slow clients are dropped instead of faulting the pipeline and the pump completes its writer on failure; profile recall under the gain lock logs the skipped gain instead of returning an error; stale direct links are cleared after the transactional swap, once the chain route is live; and the serial redaction from earlier. Your CI workflow, tests, transactional rebuild, process handling and the deck changes are in as you wrote them. Thanks again for the work.

PR #10, opened 2026-09-03
Editable mixer routing, native LSP editors and supervised daemon recovery

CarinaSchoppe:
## Summary

Adds editable application channels and output mixes, native LSP editors on the processing instance, independent per-channel DSP, and supervised daemon/plugin recovery. Also adds manual restart controls, richer local diagnostics, version-aware update notices, and installed-package acceptance on Ubuntu, Fedora and Arch.

This PR is directly from **CarinaSchoppe:main** into **emaspa:main**, includes upstream through `4e7b1ae`, and does not create or claim an upstream release. Current head: `19c6dd5` (implementation: `eda5967`; subsequent commits record acceptance and make the catalog-coalescing test deterministic).

## Mixer layout, routing and persistence

- Add, rename and delete application channels and virtual microphone outputs beyond the initial Chat/Stream pair. Stable IDs are separate from display names; existing settings retain their default layout.
- Protect hardware inputs, Monitor/Aux, and the final application channel. Deletion removes owned nodes, sends and insert references, moves affected application assignments to a fallback, and repairs enforced-default references. Profile and OpenDeck references continue using stable IDs; invalid/stale targets are rejected rather than creating ghost settings.
- Serialize structural edits with sweeps and saves. Persist successful changes before acknowledging them; clean partially built graphs and restore the previous layout if reconstruction fails. Structural edits briefly interrupt the owned audio graph.
- Edit individual channel sends from mixer cards or Flow, and assign running applications directly in Flow. State refreshes preserve open routing pickers. Requests are correlated with results; failures are visible and older daemons disable unsupported actions.
- Application audio enters a stable public sink, passes through its own stereo insert chain, then fans out into independently controlled sends. Aux In also has its own stereo insert path; output mixes retain independent processing.
- Application fan-outs are internal `Audio/Filter` nodes with explicitly configured DSP input ports, not extra selectable Pulse playback sinks. Internal stages carry role-specific descriptions and filter metadata, and are excluded from OpenXLR device pickers. Hardware meters and mix/monitor taps still require low-level routing nodes; this PR does not pretend those nodes disappear from graph inspectors.
- Quote module property lists correctly, including display names containing quotes, apostrophes and backslashes.

## Native plugin UI and DSP ownership

- **Native plugin UI** opens the installed LSP X11/XWayland editor on the same LV2 instance that processes audio. Its FFT/history/compression display is the real vendor visualization. Native float-control edits return to the daemon and persisted mixer settings.
- Generated controls remain available with unit labels, named scale-point choices, rotary keyboard/drag/wheel interaction, EQ-band bars and dynamics threshold overviews. These generated charts are labelled parameter overviews, not measured response curves.
- Each insert runs in an isolated C host with preallocated realtime buffers, atomic controls, bounded IPC/startup/UI waits and strict required-feature validation. The realtime callback performs no host-side logging, allocation, UI calls or locking.
- One-second progress heartbeats cover the host main/UI loop and active audio callback; a ten-second no-progress timeout causes supervised reconstruction with saved controls. Parent-death handling prevents orphan DSP processes/nodes. Closing an editor does not destroy processing.
- Unsupported LV2 requirements fail explicitly. Supported contract and extension boundaries are documented in `native/README.md`.

## Recovery and diagnostics

- Install systemd `Type=notify`, `WatchdogSec=60`, `Restart=always`, `RestartSec=3`. Explicit service stops stay stopped. Watchdog notifications are independent of the UI and gated by mixer/device progress; missing hardware is not considered a hang. Startup failures exit for systemd to retry.
- An always-available **Restart daemon** action in the header and Options shares asynchronous busy/error state. Service processes have bounded waits and drained stdout/stderr, so requesting recovery does not block the UI.
- Bound WebSocket connection/send and ping/pong waits; reconnect without replaying edits that may already have been applied. Coalesce notifications outside the USB lock, prevent overlapping meter/sweep callbacks and support clean shutdown with connected clients.
- Ship a PipeWire-Pulse-only `LimitNOFILE=16384` override in package/service integrations. Real testing reproduced the 1024-descriptor limit with larger layouts; raising only the OpenXLR process limit would not fix the audio server.
- Local diagnostic archives now include service state/restart/watchdog details, bounded recent PipeWire/Pulse/WirePlumber journals, build identity, bounded UI-session events and a graph summary distinguishing duplicate node names from intentional processing stages.
- Archives are owner-only, redact common home/host/USB-serial/process fields, include a privacy notice and are never automatically uploaded. Users should still inspect an archive before sharing it publicly.

## Startup updates and provenance

- Optional asynchronous GitHub check once at UI startup, with manual retry and readable change details. Audio does not depend on update-check success.
- Identify repository, version, commit and release/development-snapshot status in About and the update UI. Fork development commits are not presented as a new upstream release.
- Compare stable numeric release versions and only report development changes when remote main is actually ahead of the installed commit. Handle missing releases, malformed/oversized responses, cancellation, rate limits/network errors and duplicate requests.
- Read-only unauthenticated HTTPS requests, time and response-size limits, no redirects or auto-installation, and plain-text changelogs. Repository identity is configurable at build time and follows the repository in CI.

## Packaging and quality gates

- New Ubuntu 24.04 / Fedora 44 / Arch rolling workflow builds and installs actual DEB/RPM/Arch packages, then launches the installed daemon, native helper and Avalonia UI under an unprivileged user.
- Correct required Pulse CLI dependencies and Fedora `pipewire-utils`; discover Fedora's `/usr/lib64/lv2` bundles so LSP tests are not silently skipped. Use the bundled Inter font explicitly as default, including on systems with no system fonts.
- Add a checksum-validated local Arch build path using committed HEAD, with no automatic installation or service changes. Carry source revision into source-archive builds.
- Isolate PipeWire/Pulse/WirePlumber/D-Bus and Xvfb during package tests; hardware monitors are disabled. Local native-editor tests isolate legacy LSP configuration with bubblewrap. GitHub-hosted runners use a verified-fresh disposable profile without weakening namespace security.
- README, manual, API, architecture, native-host contract, installation guide and verification record describe behavior, migration, recovery and remaining scope.

## Verified results

Current upstream PR checks: [general CI](https://github.com/emaspa/openxlr/actions/runs/33779237338) and [package/runtime matrix](https://github.com/emaspa/openxlr/actions/runs/33779237079). The implementation passed [its fork package matrix](https://github.com/CarinaSchoppe/openxlr/actions/runs/33778085911), and the documentation follow-up passed [its complete fork matrix](https://github.com/CarinaSchoppe/openxlr/actions/runs/33778521565).

The first upstream Arch run exposed a scheduler-sensitive catalog test: four requests expired within its two-second budget. Commit `19c6dd5` replaces the arbitrary reply sleep with an explicit gate, requires all four callers to be pending before releasing the response, and checks the one-request assertion after a server acknowledgement barrier. It retains a bounded ten-second test timeout and changes no production timeout. The full local suite passed again; the linked upstream checks rerun the complete matrix.

- **113 .NET tests passed, zero failures and zero skips on each of Ubuntu, Fedora and Arch**, plus two offline Python acceptance-driver tests. Actual Avalonia controls and loopback WebSockets cover editing, routing, native-editor requests, restart/update controls and errors.
- Each distribution built and installed its package, opened a real UI window, exercised create/rename/send/delete through the installed daemon, checked saved cleanup and clean shutdown, and ran a native LSP gesture plus measured DSP regression using the installed helper.
- Local Release build/publish, .NET style/warning-level analyzer checks and whitespace validation passed; strict native C compilation and GCC static analysis passed. OpenDeck JavaScript/manifest checks passed in CI. The headless suite also passed with zero system fonts.
- The generated Arch archive's extracted executables passed runtime/native tests on CachyOS, and the installed desktop build remained running through a real service restart. The manual restart request returned in **0.01 seconds**, followed by automatic WebSocket/state recovery.
- Live GitHub check correctly recognized the fork snapshot and did not invent a newer release. A real diagnostic archive had 21 entries, parseable graph/state JSON, no duplicate OpenXLR node names, redacted home paths and mode `0600`. No diagnostics were uploaded.

The disruptive full live test on the CachyOS/PipeWire user session additionally verified:

| Scenario | Evidence |
| --- | --- |
| Application assignment and live reassignment | Actual Pulse sink-input destinations matched requested channels |
| Concurrent structural edits | Two clients created channels without losing updates |
| Compressor and bypass | Dry 0.2000, processed 0.0500, bypass 0.2000 peak |
| Independent channel effects | Processed channel 0.1000, other channel 0.2000 |
| Native Output knob | Gesture changed daemon control, persisted value and measured audio |
| EQ response | 440 Hz 0.0500; 6 kHz 0.1980 |
| Plugin host SIGKILL / SIGSTOP | Processing reconstructed with saved controls |
| Daemon SIGKILL / watchdog SIGSTOP | Layout recovered without duplicate nodes |
| Delete | Owned nodes, inserts and sends removed; assignments fell back |
| Shutdown with a connected client | 0.62 seconds in the latest full native-UI run |

The live test restores the original daemon/defaults in `finally`. A stalled shared PipeWire session initially produced no samples with both the unchanged baseline and the modified build, while private-server tests passed; restarting the audio session restored live audio. This is documented rather than misrepresented as proof that OpenXLR's watchdog repairs every PipeWire/kernel failure.

## Review and compatibility notes

- Review layout/reference transactions first (`MixerModel`, `MixerSettings`, `MixerService`), graph lifecycle second (`Mixer`, `PipeWireAdapter`), then native hosting/recovery and UI/service/update boundaries. Reproducible safe and disruptive commands are in `docs/verification.md`.
- Native UI support is currently X11/XWayland with float controls and LSP direct instance access. Arbitrary GTK/Qt UI embedding, LV2 worker/state/file extensions and atom-message transport are not advertised; broader plugin support requires extending the documented contract.
- Graph edits briefly interrupt audio. A frozen worker may take roughly 60–90 seconds to be recovered by progress expiry plus the systemd watchdog deadline. A failed shared audio server or kernel driver is a separate failure domain.
- Install the matching daemon, helper and UI; reload the user unit. Applying the Pulse descriptor-limit override requires a planned audio-server restart.
- CI validates Linux packages and private audio/UI runtime, not physical USB devices or a full systemd desktop boot in every distribution. NixOS integration was updated but not built in this matrix. This is Wave Link-style workflow improvement, not a claim of complete Wave 3/Wave Link parity or universal LV2/hardware compatibility.


CarinaSchoppe:
Now that I think about it, a tldr version would be I added a plugin-ui fixed a shit ton of bugs tested everything added what's new since update page more debugging features for us and editable sinks. 

PS: I hope I can have maintainer access as waiting for those features to be released sucks and running them just on mine isn't fair cuz they're awesome. 

Btw. Thx for reading so far. Guess I do a ton of ui feature stuff and you the underlying hardware reverse engineering. 

Cheers Carina

CarinaSchoppe:
Im done for today with the code 
@emaspa  you can merge it now, ill just fix the cli error and then everything should be fine (upgraded version to 0.1.16

CarinaSchoppe:
Hey @emaspa,
just wanted to leave a quick update on what I'm still working on before I consider my larger feature work on OpenXLR essentially complete.

My goal with the remaining work is to close the major software-side feature gaps compared to Elgato Wave Link / Wave 3, while keeping everything Linux-native and integrated into the existing OpenXLR architecture rather than building parallel systems.

The main things I'm currently working on / finishing are:

a proper many-to-many mix/output routing matrix

native Linux VST3 support alongside the existing LV2 support

Wave FX / VST Insert integration where the actual hardware/USB topology supports it

FX Select / processing-stage selection (raw, hardware FX, plugin/full FX, etc.)

plugin presets and complete effect-chain presets, including state persistence/import/export

proper plugin latency compensation

stronger plugin crash/hang isolation and fail-open recovery

an isolated LV2/VST3 scanner, including rescan and quarantine for problematic plugins

sidechain / auxiliary plugin buses

realistic yabridge compatibility for Windows VST3s exposed as VST3 plugins

more glitch-free effect-chain editing when adding/removing/reordering/bypassing plugins

per-mix FX routing, so monitoring can stay low-latency while applications/stream outputs receive the full processing chain

a proper Sound Check feature: record the microphone once, loop it through the real processing chain, and adjust/compare FX and presets while listening to the loop

the corresponding API/WebSocket/OpenAPI, persistence/profile, diagnostics, testing and documentation work

I'm also putting quite a bit of emphasis on keeping this production-quality: clean code, no dead/placeholder code, proper lifecycle/resource handling, real audio/DSP tests where appropriate, crash recovery, backwards-compatible migrations and avoiding unnecessary PipeWire graph rebuilds.

Once those parts are done and stable, I think I'll consider my major feature expansion of OpenXLR basically finished for now. At that point, from my perspective, the important software features I personally miss from the Wave Link / Wave 3 ecosystem should be covered, and I'd rather focus on maintaining, fixing and polishing what is there instead of continuously expanding the scope.

For hardware-specific work, especially reverse engineering additional Elgato devices, device-specific USB behaviour/registers and expanding physical hardware support, I'm happy to leave that primarily to you. I don't want to guess undocumented hardware behaviour or add speculative support, and that's also much closer to the original hardware-focused work you've already done on OpenXLR.

So essentially my split is:

My current focus: software mixer, routing, plugins/FX, VST3, Sound Check, API, robustness and UX.

Your territory / where I'd defer to you: deeper hardware support, additional devices and hardware-specific reverse engineering.

Of course I'm not trying to draw a hard ownership boundary — I'll still fix things and contribute where useful. I mainly wanted to give you an idea of where I'm heading with the current PR and where I intend to stop adding major new subsystems. 🙂

Thanks again for creating OpenXLR in the first place. The hardware work you've already done is what made all of this possible.

emaspa:
Thanks Carina, there is a lot of good work in here, and much of it is on the roadmap I have just published (docs/roadmap.md). I cannot merge it as one unit though: it carries six independent changes, several of which alter hardware-verified behaviour, and it conflicts with 0.1.14. Could you split it? In the order I can take them:

1. The small fixes: the stream-move verification, the pw-dump multi-batch parsing, the DaemonClient hardening, the hub coalescing, the restart button and the api.md field fix. I can merge these as soon as they are separate.
2. Editable channels and mixes with the setup, channel editor and Flow editing, plus the dynamic OpenDeck lists. Two changes first: a rename must only update the node description, and creating a channel must add nodes without tearing the graph down, so running apps and the virtual mics are never dropped. I will verify the fan-out nodes and the new property quoting on the Pro before it lands.
3. The watchdog, with Restart=on-failure, a start limit, SIGTERM rather than SIGKILL, and progress marked inside long operations. The pipewire-pulse LimitNOFILE drop-in has to be opt-in or documented; the packages should not change another service for every user.
4. The native LV2 host as its own PR. It decides the project's plugin direction, so it needs its own discussion: build opt-out (the .NET build must not require a C toolchain), the PDEATHSIG-on-thread issue, and separating "UI stalled" from "audio stalled".
5. The HTTP API once Origin and Content-Type checks exist; the update check default off or as a one-time prompt, throttled.

Things I will not take: the version bump and changelog entries (I set those at release time), the knob controls window and the forced dark theme (the roadmap has a token pass first, then themes), and the CI matrix and Python tooling for now. Please drop the README credit paragraph and the fork links from this PR: the README now has a [Credits section](https://github.com/emaspa/openxlr/blob/main/README.md#credits) and you are already in it, for #4 and for this work in progress, so nothing is lost by dropping them. On maintainer access: not yet. Land a few of the split PRs against the roadmap and we talk again. Thank you for the effort, genuinely; the split is what makes it usable.

CarinaSchoppe:
The first review block is now split into independent PRs based on current upstream 0.1.18:

- #12: stream-move verification.
- #13: pw-dump multi-batch parsing.
- #14: DaemonClient hardening.
- #15: hub coalescing.
- #16: nonblocking restart button.
- #17: api.md command field correction.

All six pass the existing upstream CI and are conflict-free against main. They
also merge together locally with an 80-test passing suite. None includes the
version bump, release changelogs, README credit paragraph/fork links, forced
theme, knob window, or new CI matrix/Python tooling.

The watchdog is separate in draft #18 with Restart=on-failure, a start limit,
SIGTERM, intermediate progress reporting and no pipewire-pulse limit override.
Its body distinguishes tested systemd policy and notification behavior from the
remaining real-daemon/hardware acceptance. It is not marked ready ahead of the
requested review order.

Editable layout/Flow/OpenDeck, the optional native LV2 host and the API/update
notice still need their separate upstream ports. In particular, I am not carrying
the old whole-graph rename/create behavior forward, nor replacing the newer
hardware fixes or API validation. The newer roadmap's authentication requirement
will be included in the API work. This comment is a progress update, not a claim
that all of the old combined PR has already been ported.

CarinaSchoppe:
Split progress against current `main`:

- the six small fixes are merged as #12–#17
- watchdog/lifecycle work is isolated in #18; the `READY=1` review issue is fixed by sending readiness from `ApplicationStarted`, with the timer retained for startup timeout extension and watchdog heartbeats
- native LV2 UI hosting is isolated in #19; UI-required-feature validation and live native-host/bypass/error gating are now included
- editable application channels and virtual-output mixes are now isolated in #22, including incremental channel creation, description-only renames, the setup/Flow editors, acknowledged durable commands, safe property quoting, and dynamic OpenDeck choices

#18, #19, and #22 remain drafts only for the real-device/PipeWire/Nix/packaged acceptance runs called out in their descriptions. The HTTP/update work, rejected visual knob/theme changes, CI/Python changes, and README credit/fork-link changes from this closed PR are not mixed into #22 and will stay separate if continued.

PR #18, opened 2026-09-05
Watchdog: progress-gated recovery and graceful restart

CarinaSchoppe:
## Scope

Watchdog portion split from #10 and rebased onto current upstream main (0.1.20).
Kept as a draft because the remaining hardware/audio/package acceptance should
not be implied by the automated coverage.

- systemd notify with a 60-second watchdog and progress-gated heartbeats.
- `READY=1` is dispatched directly from the `ApplicationStarted` event path;
  it no longer waits for the first 20-second watchdog timer tick.
- Progress after completed PipeWire helper operations, including failures, and
  between device-loop stages. Long graph operations report intermediate progress.
- Startup timeout extensions only while workers make progress. The timer remains
  responsible for `WATCHDOG=1` and startup `EXTEND_TIMEOUT_USEC` notifications.
- Restart=on-failure, three starts in five minutes, WatchdogSignal=SIGTERM and
  the existing graceful shutdown timeout. Packaged, generated source-install
  and Nix unit definitions follow the same policy.
- A failed initial graph build disables the mixer progress requirement while
  device control continues. Audio-server absence is not itself a watchdog fault.
- No pipewire-pulse LimitNOFILE override, global service modification, native host,
  UI redesign, HTTP API, version bump or CI/Python tooling change.

## Verification

- Release solution build/test with warnings as errors: 96 tests passed, zero
  warnings/errors. This includes a real Unix datagram test proving that
  `READY=1` is sent as soon as the registered ApplicationStarted token fires.
- Real isolated transient systemd user-service test: withheld heartbeat caused
  SIGTERM, the test process handled SIGTERM and exited cleanly, restarts occurred,
  and systemd stopped retrying with start-limit-hit after the third attempt.
  This used shortened timeouts and a disposable test process, not the user's
  installed OpenXLR daemon. The temporary service failure state was cleared.
- systemd-analyze parsed the repository's source-install unit but reported the
  expected missing executable at its example %h/openxlr path.

## Remaining before ready

Exercise the new daemon itself through prolonged graph construction, startup
without PipeWire, audio-server loss, deliberate worker stalls and graceful
recovery on the target hardware. The transient-service test proves the chosen
systemd policy, not OpenXLR hardware teardown. Nix evaluation and packaged runtime
acceptance are also pending. Existing hardware-verified routing is not replaced.

### DIFF
--- a/docs/install-from-source.md
+++ b/docs/install-from-source.md
@@ -117,6 +117,22 @@ systemctl --user enable --now openxlr-daemon.service
 journalctl --user -u openxlr-daemon.service -f   # watch it come up
 ```
 
+The supplied unit uses systemd notifications with a 60-second watchdog.
+Heartbeats require recent device and mixer progress, including completed steps
+inside graph operations. Failed polls still count as progress; a missing audio
+server leaves device control running rather than causing a restart loop.
+Startup timeout extensions are sent only while the workers make progress.
+
+On a watchdog timeout systemd sends SIGTERM, allowing normal graph teardown.
+Failed starts are limited to three attempts in five minutes. After fixing a
+persistent failure, use `systemctl --user reset-failed openxlr-daemon` followed
+by `systemctl --user start openxlr-daemon`. Manual launches without
+`NOTIFY_SOCKET` do not enable the watchdog.
+
+OpenXLR does not install a file-descriptor-limit override for pipewire-pulse.
+If its journal reports exhausted descriptors, inspect that service's limits
+and configure a user override explicitly for your environment.
+
 
 ## 7. OpenDeck plugin (optional)
 
--- a/packaging/nix/module.nix
+++ b/packaging/nix/module.nix
@@ -50,6 +50,10 @@ in
       description = "OpenXLR audio daemon";
       after = [ "pipewire-pulse.service" "wireplumber.service" ];
       wantedBy = [ "default.target" ];
+      unitConfig = {
+        StartLimitIntervalSec = 300;
+        StartLimitBurst = 3;
+      };
       environment = {
         OPENXLR_BUILD_MIXER = "1";
         # Plugins load inside pw-cli, a child of the daemon, so this one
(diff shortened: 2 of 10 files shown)


emaspa:
One thing to change before this leaves draft, from reading the watchdog loop: the first pass of the loop runs before ApplicationStarted has fired (hosted services start ahead of it), so READY=1 only goes out on the next timer tick, one third of the 60 s watchdog, about 20 s after a normal start. With Type=notify, `systemctl restart` blocks until READY, and the window's restart helper gives up after 15 s (UiSettings.Systemctl), so a healthy restart would show "Restart failed" on both restart buttons, the banner's and the new header one from #16.

Register on ApplicationStarted and send READY=1 there; keep the timer for WATCHDOG=1 (and the EXTEND_TIMEOUT_USEC path during startup). With that, and the acceptance runs you list (target hardware, PipeWire loss, worker stall, Nix, packaged runtime), I can take it after the six fixes and #16 are in.

CarinaSchoppe:
Addressed in `55d81f0` and rebased onto current `main`.

`ServiceWatchdog` now registers against `ApplicationStarted` before entering the watchdog loop. That event path sends `READY=1` immediately and independently of the periodic timer; the timer remains responsible for startup `EXTEND_TIMEOUT_USEC` and post-start `WATCHDOG=1`. Failed readiness sends are still retried.

I added a real Unix-datagram regression test which starts the notifier, confirms it is waiting, fires the application-started token, and receives `READY=1` without advancing a watchdog tick. The current upstream CI passes (96 tests locally, Release/warnings-as-errors).

I have kept the PR in draft because the target-hardware, PipeWire-loss/stall, Nix, and packaged-runtime acceptance listed in the PR is still outstanding; this update only claims the review fix and automated coverage.

CarinaSchoppe:
Additional runtime acceptance against the actual split daemon (`55d81f0`): launched with the mixer requested but no PipeWire server, in private network/PID/device/sysfs namespaces and temporary configuration. It degraded to device-control mode, sent `READY=1` after 0.19 seconds, and continued sending six watchdog notifications over the following ten seconds with a shortened six-second watchdog interval. Sending SIGTERM to the daemon itself produced a clean exit in 0.06 seconds.

This validates actual daemon readiness and continued worker progress in the no-audio-server startup case. It does not replace systemd enforcement tests or the remaining live graph/server-loss/worker-stall, target hardware, Nix and package acceptance. Those still keep the PR in draft.

PR #19, opened 2026-09-05
Native LV2 host: optional build and live plugin editors

CarinaSchoppe:
## Scope

Optional native LV2 host, split from #10 and based on main `9f60f5b` after the live channel creation and order PRs. No layout editor, VST3, themes, release changes or CI tooling.

- The ordinary .NET build does not invoke a C compiler. Opt in at build time with `EnableNativeLv2Host=true`.
- Native hosting is also an explicit **per-insert** choice: `nativeHost` defaults to false. Installing the helper does not switch existing settings/profiles away from PipeWire filter-chain. The control window offers a Native host toggle; disabling it returns to filter-chain. Changing the choice rebuilds the chain and may briefly interrupt audio.
- Explicit native selections without a usable helper/editor report an error, not a silent host substitution. The UI preserves the choice and allows it to be disabled if the helper becomes unavailable.
- The catalog checks the required features of both DSP and the exact selected X11 UI. Editor availability also requires an opted-in, non-bypassed, error-free insert and a live native process.
- Native editors control the same instance that processes audio. Controls return through normal settings persistence. Mixed chains retain filter-chain for all non-opted-in stages.
- Parent-owned pipe EOF/HUP controls helper lifetime. Separate audio/UI progress avoids treating a stalled editor alone as stalled DSP.

## Verification

- 217 Release tests pass, including old insert JSON defaulting to filter-chain, persisted host choices, UI payload preservation/recovery, feature gating and editor liveness. The existing upstream Lv2BundleTests xUnit2013 warning is not suppressed.
- Explicit native-enabled build and C11 helper compilation with C warnings as errors pass.
- Private PipeWire/Pulse/WirePlumber acceptance with Xvfb: an installed helper leaves a legacy insert in filter-chain; two instances of the same plugin in one chain honor different explicit host choices.
- The helper survives its creating managed Thread exiting. Mixed native/filter-chain build, dead-stage detection, cleanup and rebuild pass. The actual Mixer path passes real editor gesture, saved control values and reconstruction after its native process is killed.
- Earlier isolated runs verified continuing audio during a deliberately stalled UI thread, bounded parent-pipe cleanup, and the published optional helper. Those measurements are documented in the earlier comments, not claimed as freshly repeated for every code change.

Acceptance tools remain outside the PR. No seamless swapping is claimed.

## Maintainer sequencing

The per-insert opt-in prerequisite from the roadmap is implemented. The explicit hold until the remaining editable-layout work settles is still respected. Distribution-specific helper packaging is also still outstanding before shipping; this PR does not claim to complete it. Worker/state support, VST3/CLAP, presets and latency compensation remain separate work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,14 @@ jobs:
           python3 tools/check-openapi.py docs/openapi-v1.json
           node --test plugin/tests/*.test.mjs
 
+      - name: Optional native LV2 host compiles
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            libpipewire-0.3-dev liblilv-dev lv2-dev libx11-dev
+          make -C native
+          test -x native/openxlr-lv2-host
+
       - name: Lint shell scripts
         run: shellcheck --severity=error tools/check-version.sh tools/check-locked-restore.sh packaging/ppa/make-source.sh
 
--- a/.github/workflows/release-deb.yml
+++ b/.github/workflows/release-deb.yml
@@ -21,7 +21,8 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends \
-            build-essential debhelper dotnet-sdk-10.0 lintian
+            build-essential debhelper dotnet-sdk-10.0 lintian \
+            pkg-config libpipewire-0.3-dev liblilv-dev lv2-dev libx11-dev
 
       - name: Versions agree everywhere
         run: tools/check-version.sh
@@ -48,6 +49,7 @@ jobs:
           for f in usr/lib/systemd/user/openxlr-daemon.service \
                    usr/lib/systemd/user/pipewire-pulse.service.d/openxlr.conf \
                    usr/lib/udev/rules.d/70-openxlr.rules \
+                   usr/lib/openxlr/daemon/openxlr-lv2-host \
                    usr/bin/openxlr usr/bin/openxlr-daemon; do
             grep -q " ./$f\$" contents.txt || { echo "missing from the package: $f"; exit 1; }
           done
--- a/.github/workflows/release-rpm.yml
+++ b/.github/workflows/release-rpm.yml
@@ -20,7 +20,8 @@ jobs:
       - name: Install build dependencies
         run: |
           dnf install -y --setopt=install_weak_deps=False \
-            rpm-build rpmlint git-core dotnet-sdk-10.0 systemd-rpm-macros
+            rpm-build rpmlint git-core dotnet-sdk-10.0 systemd-rpm-macros \
+            gcc make pkgconf-pkg-config pipewire-devel lilv-devel lv2-devel libX11-devel
 
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
@@ -54,6 +55,7 @@ jobs:
           rpm -q openxlr
           test -x /usr/bin/openxlr && test -x /usr/bin/openxlr-daemon
           test -x /usr/lib/openxlr/daemon/OpenXLR.Daemon
+          test -x /usr/lib/openxlr/daemon/openxlr-lv2-host
           test -f /usr/lib/udev/rules.d/70-openxlr.rules
           test -f /usr/share/wireplumber/wireplumber.conf.d/50-xlr-dock-capture-hold.conf
           test -f /usr/lib/systemd/user/openxlr-daemon.service
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ fork/
 # build output
 src/*/bin/
 src/*/obj/
+native/openxlr-lv2-host
 
 # Local working material, never shipped: personal Windows exports, USB capture
 # analysis, third-party clones, caches.
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ fork/
 src/*/bin/
 src/*/obj/
 native/openxlr-lv2-host
+# nix build output symlink
+result
 
 # Local working material, never shipped: personal Windows exports, USB capture
 # analysis, third-party clones, caches.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -44,6 +44,7 @@ tools/check-version.sh
 tools/check-locked-restore.sh
 tools/check-openapi.py docs/openapi-v1.json
 tools/check-spec.py packaging/rpm/openxlr.spec
+make -C native  # optional LV2 host; needs the PipeWire, lilv, LV2 and X11 headers
 ```
 
 CI runs exactly these; the build treats warnings as errors, so a build
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,7 @@ tools/check-version.sh                          # the five version locations agr
 tools/check-locked-restore.sh                   # every packaging path restores locked
 tools/check-openapi.py docs/openapi-v1.json     # the HTTP API document keeps its shape
 tools/check-spec.py packaging/rpm/openxlr.spec  # every installed file is in %files
+make -C native  # optional LV2 host; needs the PipeWire, lilv, LV2 and X11 headers
 ```
 
 If you add or change a NuGet package, regenerate the lock files with a
--- a/README.md
+++ b/README.md
@@ -259,9 +259,10 @@ Code:
   graph changes, the bounded send queue, diagnostics redaction, systemd
   sandboxing, the test project and CI), the progress-gated watchdog, the
   update notice, the versioned HTTP API, the OpenDeck choices generated
-  from daemon state, and the first pieces of the editable layout: its
-  saved format, live channel creation and the saved order. Her native LV2
-  plugin editors are in review.
+  from daemon state, the first pieces of the editable layout (its saved
+  format, live channel creation and the saved order), and the optional
+  native LV2 host that opens a plugin's own editor on the instance
+  processing your audio.
 - [Michael Brooks](https://github.com/Michael-Brooks): the stream-sweep
   starvation fix ([#7](https://github.com/emaspa/openxlr/pull/7)) and the
   diagnosis that led to it.
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,9 @@ Source: openxlr
 Section: sound
 Priority: optional
 Maintainer: Emanuele Sparvoli <sparvoli@gmail.com>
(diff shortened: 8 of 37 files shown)


emaspa:
Two gaps in the editor-availability logic, from reading the diff; the rest waits for the acceptance you list.

1. `PluginInfo.NativeEditorAvailable` checks the plugin's DSP required features, but the X11 UI the helper would load declares its own required features and the helper only checks those at click time. LV2 says a host must not load a UI whose required features it lacks, so the catalog should evaluate the selected X11 UI's features and hide the button when they are not provided.
2. The window's `NativeEditorAvailable` is bypass plus the catalog flag; it ignores the insert's live error and whether a host process is actually running for that insert. A chain that failed to build keeps an enabled "Plugin UI" button that can only return an error. Expose host liveness in the insert status and disable the button when there is no running host.

The optional build (EnableNativeLv2Host off by default) is the right shape; thanks for that.

CarinaSchoppe:
Addressed both gaps in `889ea4c` and rebased onto current `main`.

1. The catalog now reads `lv2:requiredFeature` from the exact X11 UI the helper selects (the first X11 UI in lilv catalog order), using the same supported UI-feature set as the helper. An editor with an unsupported UI requirement is no longer advertised, so its button is hidden before click time.
2. `InsertStatus` now carries `nativeHostRunning`, derived from the insert stage's live native process. The window separates editor support from live availability: unsupported editors are hidden; supported editors are disabled while bypassed, in error, or without a running native host.

The regression coverage includes the UI feature contract, serialized liveness status, and button gating for missing host, live host, chain error, and bypass. Current upstream CI passes; locally the default Release suite is 100/100 and the explicit native-enabled build compiles the C11 helper with warnings as errors and finishes with zero warnings/errors.

The PR remains a draft only for the real editor/mixed-chain/stall/package acceptance already listed in its description.

CarinaSchoppe:
Additional acceptance against the current split helper (`889ea4c`), using private PipeWire/Pulse/WirePlumber and Xvfb:

- A real LSP compressor editor gesture changed the control callback and measured audio from the same processing instance.
- Measured gain: 0.25 → 0.0500 peak, 0.5 → 0.1000 peak; compressed peak 0.1489. Graceful host shutdown passed.
- A test-only signal handler deliberately held the main/UI thread for more than 12 seconds while the separate audio thread continued. Audio remained at 0.1489 and audio heartbeats continued while UI heartbeats stopped.
- Closing the parent-owned input pipe while that main/UI thread was still blocked terminated the helper within five seconds.

The test scripts and signal-injection helper remain outside this PR. These results narrow the remaining acceptance; they do not yet cover daemon-managed mixed-chain reconstruction, launch from an exiting managed thread, or optional package integration. Keeping draft for those remaining items.

CarinaSchoppe:
The remaining in-scope runtime checks now pass against this split: an exiting managed Thread, a real mixed native/filter-chain with editor-to-settings persistence, and Mixer-driven reconstruction after killing the native stage. The published optional helper also passed the real native gesture and measured DSP checks. Combined with the previously reported stalled-UI/continuing-audio and parent-pipe cleanup runs, this is ready for the host-direction review. I updated the description with the precise results and boundaries and am taking it out of draft.

emaspa:
Thanks for closing the two gaps and for the acceptance runs, the split is what I asked for and it merges onto main with one line of conflict in docs/api.md. I am going to hold it for now rather than merge it. Two things shipped this week that I want settled first: the Monitor A and B work with per-output feeds from #21, and the watchdog and update notice from #18 and #23, which change how the daemon runs under systemd. I would rather have those stable in users' hands, and the editable mixer layout from #22 rebuilt on that model, before the plugin side grows a second host. When the mixer side is where it should be, this PR is the starting point for native editors, and the packaging plan is straightforward since every library the helper links already ships with the daemon. Leaving it open.

CarinaSchoppe:
Updated the branch with main 4e56d4d (0.1.23) in merge commit 2d2e371. The docs/api.md conflict is resolved by retaining both showInsertUi and the current ignore-routing documentation. No monitor-feed or service behavior was reverted.

Verification on the merged tree: all 168 Release tests passed; the native-enabled solution build succeeded. I also rebuilt and reran the isolated managed runtime acceptance: host survival after the creating managed Thread exits, mixed native/filter-chain cleanup and rebuild, and the production Mixer path with a real editor gesture, settings persistence and reconstruction after the native stage is killed all passed.

Your hold remains in place: this update only keeps the accepted starting point compatible with current main. It does not ask to move native hosting ahead of the editable mixer work. The description now also makes the roadmap requirements for per-insert opt-in and packaging explicit.

CarinaSchoppe:
Rebased onto current main (89e8c9f, including 0.1.24 and the merged HTTP/OpenDeck work), retaining the two feature commits. Both are now signed and verified by GitHub: efede9d and 43088b7. Each signed commit tree was compared with its corresponding local rebased commit and is identical. The previous history is preserved on an archive branch. All 185 Release tests pass on this base. This fixes the signature requirement without changing the requested hold behind the mixer work.

emaspa:
@CarinaSchoppe do you use discord? would be great if you could join the new official server

CarinaSchoppe:
Updated onto main 9f60f5b after #34 and #35 merged. The two native feature commits remain separate and are verified: 5551019 and 6df8345. All 214 Release tests pass on the combined tree; the helper rebuilt with C warnings as errors and the isolated managed-thread/mixed-chain/editor-persistence/crash-reconstruction acceptance passed again. GitHub reports the branch conflict-free; fresh CI/CodeQL is running. I am preserving your explicit hold behind the remaining editable-layout work rather than interpreting technical mergeability as lifting it.

CarinaSchoppe:
Implemented the roadmap prerequisite for explicit per-insert host selection. nativeHost defaults to false, including old settings/profiles; the controls window now offers the choice and preserves it in full-chain edits. An installed helper no longer silently switches existing chains. Explicit unavailable native selections report an error and can be disabled in the UI.

All 217 Release tests pass. The private PipeWire acceptance now additionally verifies that an old insert stays in filter-chain with the helper installed, and two instances of the same plugin in one chain respect different host choices. Managed-thread lifetime, actual editor gesture/persistence and crash reconstruction still pass. The new commit is signed. This closes the per-insert-choice item, not the remaining packaging requirement.

emaspa:
Thanks for finishing this. I read through the branch on current main and built the helper here with the Makefile (gcc, pipewire 1.6.8, lilv 0.28); it compiles clean with -Werror and links only libpipewire, liblilv and libX11.

The shape is right: default build untouched, per-insert opt-in that defaults to off, the safety chain kept in front, and the helper exiting on its own when the daemon goes away. Four things I would like fixed before it merges:

1. Show/set round trips run under the mixer lock. Mixer.ShowInsertUi and SetInsertParam hold _gate while NativePluginHost waits up to 2 s to send and 5 s for the reply. A client repeating showInsertUi stalls every other command and the device sweep for that long. Do the round trip outside the lock, or cap it at about a second.

2. No X error handlers in the helper. Without XSetErrorHandler/XSetIOErrorHandler a toolkit protocol error or an XWayland restart runs Xlib's default handler, which exits the process and takes the audio instance with it. The chain rebuilds on the next sweep, so it is a dropout rather than a daemon crash, but it breaks the "UI trouble never touches DSP" promise. The handlers should close the window and keep processing.

3. DISPLAY and XAUTHORITY under the packaged unit. XOpenDisplay(NULL) inherits the daemon's environment. The shipped unit starts at default.target before the desktop's import-environment, PrivateTmp hides /tmp/.X11-unix, and GDM/SDDM cookies are not in ~/.Xauthority. Under Xvfb this is invisible; on a real install the button will most likely answer "ui unavailable". Either document the import-environment step in native/README.md or have the daemon locate the session's display and cookie and pass them to the helper.

4. Small alignment: add lv2 to PKGS in native/Makefile (it works today only because lilv-0.pc pulls it in), and align MAX_PORTS with the catalog's MaxPorts (2048 vs 4096) so a plugin cannot pass validation and then fail at build.

Tests are hermetic and cover validation, defaults, JSON and UI gating. If you can make the helper path injectable, a test around the ready/heartbeat/timeout protocol with a fake helper would be worth having; the USB helper tests in UsbHelperTests.cs use a Python fake for the same purpose.

On timing: the hold stays for now. Main already carries the HTTP API, the OpenDeck choices and the layout work since 0.1.24, and I want rename, delete and the desktop editor in and tested on real hardware before adding a native process and a set of new build dependencies to every package. Once the mixer work is mature this goes in next. Keeping it rebased on main in the meantime is appreciated.

emaspa:
Rebased onto main and finished here, so this branch now carries your four
commits plus five of mine. The previous head was 7bc19c7.

**An editor failure no longer takes the audio with it.** Xlib exits the
process on both of its error paths, so a protocol error from a plugin's
editor, or a display that goes away, killed the LV2 instance that was
processing audio. A protocol error is now logged and the editor carries on.
A lost connection drops the editor and leaves the plugin running, and the
next request opens a fresh one. Verified with x42 darc on a mix by asking
the X server to close the helper's own connection: the same process kept
its ports, its meters and its heartbeat, and the editor reopened. A build
with the two handlers stripped died the instant the connection closed.

**Liveness.** The heartbeat no longer depends on the graph having audio to
process, so an idle mix cannot make a healthy host look dead. A frozen
editor loop is reported on the insert as frozen controls rather than taken
for a dead process, and the editor timeouts now say what happened to the
audio instead of surfacing a cancelled task.

**Bounded restarts.** A plugin that dies as soon as it starts was rebuilt on
every sweep, and rebuilding an input chain interrupts the microphone. A
chain that fails three times within five minutes is left off with the reason
on the insert; changing or bypassing it starts over. Editor changes are now
drained before the healing pass, so a chain that is about to be rebuilt
comes back with the values its editor last showed.

**Packaging.** The helper is in every package. It links libX11, lilv and
PipeWire, which were already runtime dependencies everywhere, so only the
headers and the opt-in flag are new. Verified by building for real on all
four channels: Fedora 44 through the release workflow, with rpmlint and an
install check; Ubuntu noble the same way, with lintian; Arch with makepkg
here; and Nix, whose dotnet builder needed gcc and pkg-config of its own.
Both release workflows now assert the helper is in the package and
runnable, so this cannot quietly regress.

**Session setup.** The PrivateTmp override is out of native/README.md. The
editor opens from the packaged unit exactly as it ships, because Xlib
reaches XWayland through the abstract socket, so the sandbox needs no hole.
The import of DISPLAY and XAUTHORITY is still documented, in the manual at
3.12 along with the rest of the feature.

249 tests pass. Worker, state, presets, VST3 and CLAP remain separate work,
as you scoped them.

PR #23, opened 2026-09-05
Add opt-in throttled update notices

CarinaSchoppe:
## Summary

This extracts the update-notice part of #10 into a focused pull request.

- keep update checks disabled by default and require an explicit opt-in
- check at most once per day, with a separate manual **Check now** action
- query only the official `emaspa/openxlr` GitHub releases endpoint
- accept stable releases only and compare numeric semantic versions
- never download or install anything automatically
- bound the response size and timeout, reject redirects, and construct the release link locally
- render release notes as plain text and allow a release to be dismissed

## Verification

- `dotnet test OpenXLR.slnx -c Release` — 95/95 passed
- `git diff --check`
- tests cover default-off behavior, daily throttling, semantic version ordering, safe release URLs/plain-text notes, ignored drafts/prereleases, oversized responses, and a stalled response body after headers

This intentionally does not include the authenticated HTTP API or any unrelated mixer/UI work from #10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,9 @@ Collect diagnostics).
   and source once a second, and serves a WebSocket API on
   127.0.0.1:37890. The UI has a routing graph view, a tray icon and a
   diagnostics archive exporter.
+- **Optional update notice**: the UI can check the upstream GitHub release
+  feed for a newer stable release. Startup checks are off by default and,
+  when enabled, run at most once per day. Nothing is installed automatically.
 
 The full feature list, area by area: [docs/features.md](docs/features.md).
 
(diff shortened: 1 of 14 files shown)


CarinaSchoppe:
Found and fixed one more timeout issue during the follow-up review: `ResponseHeadersRead` only lets `HttpClient.Timeout` cover the response headers. Commit `c3d6adb` adds a linked deadline covering the entire body read as well. An offline regression test returns headers immediately and then stalls the body; cancellation now ends that read within the same deadline.

Release tests pass 95/95 with warnings treated as errors, and upstream CI passes. This PR is ready for review.

PR #24, opened 2026-09-05
Add versioned HTTP API with the existing session token

CarinaSchoppe:
Adds versioned HTTP access to the existing command dispatcher on current main. HTTP clients use the daemon's existing per-session token in a Bearer header; the window, OpenDeck and `/ws` retain their current authentication behavior.

- state, plugin catalog and command endpoints, plus `/api/v1/events` using the existing authenticated WebSocket protocol
- Origin and UTF-8 JSON Content-Type checks; 64 KiB limit including chunked bodies, body-read deadline, command budget and no waiting mutation queue
- shared validation, command dispatch and broadcasts, including current Monitor A/B feed commands
- HTTP/OpenAPI documentation with token paths, failure statuses and retry semantics

Verification: 158 Release tests passed, including real HTTP requests for authentication, foreign Origin, content type, oversized body, rejected commands and successful dispatch; `git diff --check` passed. Current upstream tests emit existing CA1416 platform warnings in ApiTokenTests/PrivateFilesTests; this PR does not suppress them.

This is the HTTP API portion of #10, rebuilt on upstream's authentication and socket hardening. It adds no second credential system, layout change, native host, version change or CI tooling.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,5 +1,8 @@
 # WebSocket API
 
+The [versioned local HTTP API](http-api.md) provides request/response access to
+the same commands and uses the same per-session credential.
+
 The daemon serves `ws://127.0.0.1:37890/ws` on the loopback interface
 only. A handshake that carries a browser `Origin` header from anywhere
 but localhost is refused with 403, so a web page you happen to visit
--- a/docs/http-api.md
+++ b/docs/http-api.md
@@ -1,0 +1,45 @@
+# Local HTTP API v1
+
+The daemon listens on `127.0.0.1:37890`. HTTP requests use the same per-session
+token as the existing WebSocket clients, presented as `Authorization: Bearer
+<token>`. Read it from `$XDG_RUNTIME_DIR/openxlr/token`, or from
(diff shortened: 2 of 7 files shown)


emaspa:
Ready to merge once the commit is signed; see the note on #25 for the two-minute SSH signing setup. I merge both as soon as they show as verified.

emaspa:
Rebased onto main (0.1.24) and pushed. One change of substance: the endpoints now read the daemon's token per request instead of capturing it when mapped, because since 0.1.24 the token is published only once the port is bound, after mapping, so the branch as it stood refused every HTTP request with 401. The test publishes a token the way the daemon does. The rebase re-signed your commit with my key; authorship is unchanged. Merging once CI is green.

CarinaSchoppe:
Thanks for rebasing, fixing token-publication timing and signing the commits. I have preserved your merged version. I also inspected the new CodeQL result on this PR: it flagged the path-dependent authentication exception at ApiEndpoints.cs:50. Follow-up #32 binds HTTP authentication to the resolved endpoint group, keeping events separately first-frame authenticated and Origin-checked. It includes real HTTP regression cases and passes all 167 Release tests; no CodeQL suppression or protection changes.

PR #22, opened 2026-09-05
Make mixer channels and outputs editable

CarinaSchoppe:
> **Runtime blocker:** channel/output rename currently fails because the adapter uses unsupported pactl commands. Incremental creation preserves existing node identities in the private PipeWire test. Rename still needs a working owner-side metadata update before this PR can leave draft.

## Summary

This extracts the editable mixer-layout work from the closed all-in-one #10 onto current `main`.

- persist user-managed application channels and virtual-output mixes with stable IDs, while Monitor, Aux, and hardware inputs remain structural
- add the Channels & outputs dialog, live app routing in Flow, and a focused channel-send editor opened from Flow
- create application channels incrementally: only the new public sink and internal fan-out are added, so existing application sinks and virtual microphones keep their PipeWire nodes
- quote creation-time module properties for names containing spaces, quotes, apostrophes, and backslashes; the intended description-only rename path is currently blocked as noted above
- acknowledge layout commands only after the new layout is saved, with serialized edits and an authoritative state push before `commandResult`
- build OpenDeck mix/send choices from live daemon state, keeping stable IDs in saved actions and current editable names in labels
- keep delete-channel/delete-output and output creation rollback-safe; these matrix row/column changes may still briefly rebuild the owned graph

## Verification

- `dotnet test src/OpenXLR.slnx -c Release --no-restore` — 98/98 passed
- `node --check plugin/com.emaspa.openxlr.sdPlugin/plugin.mjs`
- `node --test plugin/tests/*.test.mjs` — 2/2 passed
- `git diff --check`

## Acceptance still needed before leaving draft

- on a real PipeWire desktop, add an application channel while apps are playing and OBS/Discord hold existing OpenXLR virtual microphones; confirm those existing endpoints never disappear and audio is not dropped
- rename channels and outputs using ordinary and quoted names; confirm only labels change and saved app/profile/OpenDeck references continue to resolve
- create/delete an output and delete a channel; confirm rollback and fallback routing behavior
- exercise the packaged OpenDeck property inspectors against a changed layout

This intentionally excludes the watchdog work (#18), native LV2 editor hosting (#19), the HTTP API, visual knob/theme changes, CI/Python changes, and README credit/fork-link changes from #10.

PR #25, opened 2026-09-05
Generate OpenDeck mixer choices from daemon state

CarinaSchoppe:
Generates OpenDeck channel/mix choices and labels from the current daemon state, as the standalone OpenDeck portion requested in #22. Built directly on current main's Monitor A/B/A+B feed model.

- Stable existing targets remain unchanged: mix masters/mutes, send levels/mutes, hardware controls, monitor outputs and feed switching.
- Open property inspectors receive changed layout lists; rename, insertion and deletion update choices without rewriting saved action settings.
- Missing saved mixer targets are shown as unavailable instead of silently selecting a different action.
- Monitor A and B retain distinct dial labels. Feed groups, feed cycling, authentication, profiles and insert controls retain their current implementations.

Verification: five Node tests cover stable IDs/current names, deleted entries, selection preservation across state refreshes, both monitor mixes, and the actual plugin's property-inspector subscriptions and A/B/A+B feed commands. Plugin/inspector JavaScript syntax and `git diff --check` pass. GitHub CI passes on `93ef59f`.

This is independently mergeable before the editable-layout backend. It adds no daemon, routing, desktop-theme or release changes.

### DIFF
--- a/plugin/com.emaspa.openxlr.sdPlugin/layout-choices.mjs
+++ b/plugin/com.emaspa.openxlr.sdPlugin/layout-choices.mjs
@@ -1,0 +1,68 @@
+// Pure helpers for the editable mixer layout. Targets always carry stable
+// daemon ids; labels come from the latest state so renames do not invalidate
+// existing OpenDeck actions.
+
+const LEGACY_CHANNELS = {
+  xlr1: "XLR 1", xlr2: "XLR 2", aux: "Aux In", game: "Game",
+  music: "Music", browser: "Browser", system: "System",
+  voicechat: "Voice Chat", sfx: "SFX",
+};
+const LEGACY_MIXES = { monitor: "Monitor A", monitor2: "Monitor B", stream: "Stream", chat: "Chat", auxout: "Aux" };
+
+export function channelName(mixer, id) {
+  return mixer?.channels?.find((channel) => channel.id === id)?.name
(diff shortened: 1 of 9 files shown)


emaspa:
Both #24 and #25 are ready. I pushed one small commit on this branch (3f4560a) that keeps the generated groups in the picker where the static ones were: a hidden anchor in the two inspector pages and an insert at that anchor instead of an append, with the order asserted in the test.

One thing before merging: the repository now requires signed commits on every branch, and your commits on #24 and #25 are unsigned, so today they could only be squash-merged, which would flatten them. I would rather keep your history, so please re-push both branches signed. SSH signing takes a minute: add your public key to GitHub as a signing key (Settings, SSH and GPG keys, "Signing key"), then

```sh
git config --global gpg.format ssh
git config --global user.signingkey ~/.ssh/<key>.pub
git config --global commit.gpgsign true
git rebase --exec 'git commit --amend --no-edit -S' main   # re-signs every commit on the branch
git push --force-with-lease
```

Keep 3f4560a in the rebase; it is already signed. I merge both as soon as the commits show as verified.

emaspa:
Rebased onto main (0.1.24) and pushed; no content changes beyond the rebase, the Node tests pass. The rebase re-signed your commits with my key; authorship is unchanged. Merging once CI is green.

PR #35, opened 2026-09-06
Persist live channel and virtual-mix order without graph changes

CarinaSchoppe:
Ports the persistent layout-order portion of original fork commit 10f4ae0 onto current main, independently of live channel creation (#34).

Adds setLayoutOrder {channels, mixes}: complete permutations of editable application-channel and virtual-microphone IDs. Hardware inputs, Monitor A/B and Aux remain structural. Rejects missing, duplicate, unknown, oversized and structural IDs. The mixer retains existing definition objects, IDs, routes, levels and mutes; only display order changes. Save failure restores the exact prior config object and returns an error. Success broadcasts only after persistence. Ordinary fader saves retain their existing best-effort/retry behavior.

Both request fields are consumed by the production dispatcher and mixer, and deserialization/membership tests cover them. 185 Release tests pass. External private PipeWire acceptance through MixerService.Apply verifies successful persistence before the event, plus a real read-only mount failure with no success event, unchanged settings, restored order, and unchanged IDs/serials for every node and link. The commit is GitHub-signed and verified.

No desktop-editor changes, themes, release/CI modifications or general monitor-routing redesign. This completes the backend order operation only; desktop controls and other #22 live edits remain separate.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -54,6 +54,7 @@ Commands are single JSON objects with a `cmd` field:
 | `setSoftClipGuard` | `value` | software ClipGuard (post-ADC limiter at -3 dB); enabling is rejected if `swh-plugins` is unavailable, without replacing or disconnecting the live microphone route |
 | `setLevel` | `channel`, `mix`, `value` | one send fader |
 | `createChannel` | `name` | add an application channel without rebuilding existing nodes; succeeds only after settings are saved, with its generated stable ID in the next state |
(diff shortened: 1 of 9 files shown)


CarinaSchoppe:
Updated onto main after merging #34. Resolved the overlapping command registration, strict-save path and documentation while retaining both operations. The combined tree passes all 196 Release tests. Both private PipeWire acceptance runners were rebuilt against this same tree: createChannel and setLayoutOrder succeed only after persistence; actual read-only configuration fails without success events or state/graph leaks. Existing nodes and links remain stable. The replacement commit bc9a759 is signed and verified; waiting for its fresh CI/CodeQL checks before merging.

PR #45, opened 2026-09-07
Native host: the worker extension, and two fixes from using it

emaspa:
Three changes from putting the native host in front of a real plugin set.

**The worker extension.** Reverbs and convolvers hand their heavy work to
the host instead of doing it in the audio callback. The helper could not do
that, so they could not be hosted, so their editors could never open. It now
runs that work on a thread of its own, fed by two lock-free rings: requests
leave the audio callback, answers come back and are delivered before the
next run. Neither side allocates or locks.

Every plugin here that wants the worker also wants `options` and the promise
that the block length is bounded, so both come with it. The block length
already is bounded, by the check the audio callback makes.

On this desk that moves nine plugins from unhostable to hostable, including
the three Dragonfly reverbs and the x42 convolvers. Verified live: Dragonfly
Hall Reverb runs as an insert on a mix and its editor opens with its
spectrogram drawn, which is exactly the thing the worker computes. Every
plugin installed here that has an X11 editor can now be hosted.

**The helper is plumbing.** An insert in the native host registers a
PipeWire client, so it appeared in the window's application list as
something to route, beside Discord and Spotify. It now joins the daemon, the
window and the PipeWire tools in the ignore list. A remembered entry from
before the fix stays until it is forgotten from the list.

**Buttons on one side.** Native host and Manual sat on the left under the
plugin name while Plugin UI, Defaults and Bypass sat on the right. They now
share the right side in two rows.

253 tests. The C helper builds with warnings as errors, here and in CI.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -135,7 +135,8 @@ audio that is playing and its changes are saved like any other control.
 The plugin keeps processing if the editor stops answering or the X
 display goes away, and a chain whose host keeps crashing is switched off
 once it has failed three times in five minutes, with the reason on the
-insert. The manual covers the
+insert. Plugins that hand work to a background thread, which is how
+reverbs and convolvers build their impulse responses, are hosted too. The manual covers the
 setup in 3.12.
 
 The submixer can be switched off in Options. The daemon then controls
--- a/native/README.md
+++ b/native/README.md
@@ -73,8 +73,15 @@ Use the session's own value rather than a copy, and never `xhost +`.
 
 ## Scope
 
-URID map and unmap for the plugin, and the X11 editor features this host
-implements: instance access, parent, resize and the idle interface. Worker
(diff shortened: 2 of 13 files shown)


emaspa:
Pushed a fourth commit after hitting this on the desk: an insert whose native host is switched off could not be switched back on.

Two causes, both about a helper that had quietly gone. An ordinary build after a native one deleted it from the output, because the opt-in flag decided both whether it was built and whether it was copied. The flag now decides only the build; a helper that exists is copied either way. I checked all three paths, including a clean checkout with the flag, which is how the packages build.

The refusal also blamed the plugin rather than the missing helper. A plugin now carries two separate facts, whether its editor is one the host could open and whether the host is installed, and the daemon names the one that is missing. The window keeps the switch visible for a plugin it could host, so an insert cannot become impossible to change back.

PR #60, opened 2026-09-08
Devices: give a dock its own gain back even when the lock is on

emaspa:
Reported as two symptoms: after a reboot the gain reads 43 dB instead of the configured 55, and there is no microphone input at all, with replugging no help. Diagnosed from the reporter's diagnostics archive; both symptoms have one cause.

## What happens

An XLR Dock keeps no settings, so the daemon writes its last state, or its recall-on-connect profile, whenever it connects. The gain lock made that skip the gain, and the log said so:

```
recalled profile 'Carina' on connect of 0fd9:00a6
profile loaded; gain left unchanged because the gain lock is active
```

So the dock stayed at the 43 dB its firmware powers up with, while the window, the profile and the saved state all said 55. Every replug ran the same sequence.

## Why that silences the microphone

Her chain starts with an LSP gate whose threshold sits near -25 dBFS, tuned at the gain she meant to have. I rebuilt her exact chain, low cut, gate, compressor, with her own parameter values, and fed it a tone:

| input | output |
|---|---|
| -23.0 dBFS, a microphone at 55 dB gain | -25.5 dBFS, passes |
| -35.1 dBFS, the same at 43 dB gain | silence |

A gate that never opens is not a quiet microphone, it is a dead one, which is exactly what was reported.

## The fix

The lock is there to keep the gain where the user put it. Refusing to restore it produced the one outcome it exists to prevent. Settings a device is given back on connect are now written whole, gain included, whether or not the lock is on. Everything else about the lock is unchanged: a client setting the gain is still refused, and so is a profile loaded by hand.

Four regression tests at the DeviceManager seam, with a fake dock that powers up at 43 dB. The first one fails against the old rule (`Expected: [55], Actual: []`) and passes with the fix. 298 tests pass.

Documented in the gain lock's description, the hardware support table, and a troubleshooting entry for a microphone that is silent only after a reboot.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -51,7 +51,10 @@ For devices without the hardware version, the PipeWire layer provides:
 - Gain lock: the daemon rejects every gain change while the lock is set,
   from any client, and stores the lock per device in `gainlock.json`.
   Shown only for devices without a physical gain dial, which would
-  bypass it.
+  bypass it. Giving a device its own settings back when it connects is
+  not a change and goes through: a dock that forgets its settings would
+  otherwise come up at whatever gain its firmware chooses, which is the
+  one thing the lock is there to prevent.
 
 These controls appear only when the active device lacks the hardware
 version, so a signal is never filtered twice.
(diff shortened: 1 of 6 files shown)


CarinaSchoppe:
_Diagnosed from the reporter's diagnostics archive;_
The reporter had a name🫠

emaspa:
> _Diagnosed from the reporter's diagnostics archive;_ The reporter had a name🫠

Open an issue next time :D

PR #79, opened 2026-09-10
Window: defer close-to-tray and restore minimized windows

CarinaSchoppe:
Closing the mixer with close-to-tray enabled currently calls `Hide()` inside the native closing callback. Defer hiding until that callback returns, and cancel queued hides when the mixer is requested again or the application exits. Tray actions and subsequent launches share the same restore path, including restoring a minimized window.

Also allow a normal close if tray construction failed. Quit and application shutdown still close the window, and queued restore requests do not reopen a closed window.

Validation:
- Locked restore, Release build with warnings as errors, and all 333 existing .NET tests pass.
- The new isolated Xvfb window test passes separately and runs in CI. It covers repeated close/restore, minimized restoration, a pending hide followed by show or quit, application shutdown, and closing with the preference disabled.
- Plugin tests, native editor tests, and the complete fork CI pass: https://github.com/CarinaSchoppe/openxlr/actions/runs/34445816218

The reported freeze on the real Wayland compositor has not been directly reproduced. The window regression test disconnects from the desktop session bus and does not validate a real tray host. Desktop acceptance with the real tray menu and plugin editors, and hardware verification, remain outstanding.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,9 @@ jobs:
           test -x native/openxlr-lv2-host
           xvfb-run -a make -C native test-editor
 
+      - name: Tray window lifecycle
+        run: OPENXLR_TEST_DESKTOP=1 xvfb-run -a dotnet test src/OpenXLR.Tests/OpenXLR.Tests.csproj -c Release --no-build --filter FullyQualifiedName~TrayWindowTests
+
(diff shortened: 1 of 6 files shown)


CarinaSchoppe:
Replaced by #80, using a dedicated fix/tray-close branch so the independent window layout changes can be reviewed separately.

PR #83, opened 2026-09-11
Window: repair autostart only where needed and keep the mixer sizing limits

emaspa:
Follow-up to #80, #81 and #82. @CarinaSchoppe, this touches your three merges; the notes below say what changed in each and why, so nothing here should be a surprise.

## Fixes

**Turning mixer autostart off could fail forever.** `File.Delete` throws `DirectoryNotFoundException` when `~/.config/autostart` is absent, so `SetWindowAtLogin(false)` reported failure and the preference stayed on. Removing an entry that is already gone now counts as done.

**The autostart entry was written with OpenXLR's private-file rules.** `OpenXlrPaths.WriteAtomic` is right for `~/.config/openxlr`, where the AGENTS.md rule applies, and wrong for a shared directory: it set `~/.config/autostart` to 0700 and the entry to 0600, and its rename turned a symlinked entry into a regular file. Both startup writers (the desktop entry and the systemd user unit) now share one helper: temp file plus rename in the same directory, directory mode untouched, file mode from the umask, and a symlinked entry refused and reported rather than replaced.

**The startup repair rewrote too much.** It ran on every launch, replaced any `Exec` that differed from the running copy (reverting user edits and switching a packaged entry to a source build after one manual launch), recreated entries removed with a desktop tool, and missed the spec-legal `Exec = value` form, appending a duplicate key. The rules are now: an `Exec` whose program still exists is never touched; one whose program is gone gets this installation's path with every other line kept; a missing entry is recreated only when the installation path changed since OpenXLR last wrote one (recorded in `ui.json` as `autostartExecutable`). Keys are matched per the Desktop Entry spec and duplicate `Exec` keys collapse to one. The outcome is reported in Options instead of discarded.

**The tray gate was dead.** Avalonia 12.1.2 never throws from `new TrayIcon` on Linux: without a session bus the implementation is a stub that swallows every call, and with one it is an internal type whose `IsActive` only becomes true after the constructor returns. There is no reliable synchronous way to know whether a tray exists, so the `_tray is not null` gate is removed from both `Closing` and `StartsHidden`, and the manual now says what actually happens: with close-to-tray on, the close button always hides the window, and on a desktop without a tray the option should stay off.

**Rejected toggles left the checkbox wrong.** Raising the property with an unchanged value does not push anything back through the binding, so after a failed change the box showed the rejected state while the preference, `ui.json` and the hint said the opposite. Both login checkboxes now snap back through a set-and-revert, and `StartDaemonAtLogin` handles failure the same way as the window option instead of persisting `true` when `systemctl` fails or there is no daemon binary.

## Reverted sizing decisions

Two of #81's choices are reverted on purpose:

- `MinWidth="760"` is dropped. The window had no floor before, and below 760 the content was laid out at 760 inside a narrower surface, cutting off the right edge on a half-screen snap of a 1366 px laptop. The widest row, the seven input toggles, measures 584 px at a 640 px window, so no replacement floor is needed.
- `MaxWidth="1300"` is restored on the content panel. Without it every star-column fader and dropdown stretched across a maximized ultrawide, which the cap was added to prevent. Measured: 1224 px content at 1800 and 2400 px windows, against 1768 and 2368 without the cap.

The rest of #81 stays: ellipsis and tooltips on insert names, error rows, wrapping mix cards, plugin action rows.

## Tests

The new Xvfb tests now pin the behaviour they describe, each verified by reverting the fix on a scratch copy and watching the test go red:

- `TrayWindowTests` drives a real `WM_DELETE_WINDOW` ClientMessage through the X11 event loop (libX11 P/Invoke, already a dependency of Avalonia.X11), records visibility at the moment the window's own `Closing` handler returns so the deferral can fail, asserts visibility after `Close()` in the pending-hide-then-show and then-quit cases, and points the daemon client at an unused port through an internal `MainWindow(DaemonClient)` constructor so it no longer talks to a real daemon on a developer machine. The minimized-restore assertion, which cannot fail under Xvfb without a window manager, is dropped with a comment.
- `WindowLayoutTests` adds 640 px, asserts the 1300 cap above it and window-following below, requires exactly ten `insertName` labels including one inside each mix master card (the per-mix chain label gained that class), checks contents inside each card, and checks every plugin action button inside a 420 px window. `AssertInside` now translates coordinates so it can fail against a non-immediate ancestor.
- `WindowAutostartTests` goes from 17 to 29 cases: missing directory on disable, unchanged directory and file modes, symlinked entry refused, spaces around `=`, duplicate keys collapsed, existing `Exec` targets not rewritten (path, relative path, bare command on `PATH`), user-removed entry not recreated, `Exec` landing inside `[Desktop Entry]` with the old one gone, and the checkbox snap-back with a real bound `CheckBox`.

## Checks

- Locked restore; Release build with warnings as errors and the native host enabled: 0 warnings, 0 errors.
- `dotnet test`: 362 passed, 0 failed, 2 skipped (the Xvfb-gated pair, run separately below).
- `OPENXLR_TEST_DESKTOP=1 xvfb-run` tray test: passed. `OPENXLR_TEST_LAYOUT=1 xvfb-run` layout test at 2560x1440: passed.

## Not verified

- `SetDaemonAtLogin` has no test: exercising it runs `systemctl --user enable/disable` against the live session. Its failure reporting follows the code and `systemctl` exit codes.
- No desktop acceptance run: no real tray icon, no real `~/.config/autostart` write, no daemon contact. The D-Bus tray implementation was inspected by reflection, not exercised.
- No hardware involved; nothing here touches device code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,7 @@ OPENXLR_TEST_LAYOUT=1 xvfb-run -a -s '-screen 0 2560x1440x24' dotnet test src/Op
 
 The window layout test runs in its own process using X11 and isolated
 configuration, runtime and session-bus settings. It checks narrow plugin
-windows and mixer widths from 760 to 2400 logical pixels. Set
+windows and mixer widths from 640 to 2400 logical pixels. Set
 `OPENXLR_LAYOUT_ARTIFACTS` to a directory to save rendered previews.
 
 If you add or change a NuGet package, regenerate the lock files with a
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -109,12 +109,15 @@ on the original Wave XLR. See [hardware support](hardware-support.md).
 <a name="tasks"></a>
 ## 3. Tasks
 
-The mixer cards fill the window width when it is enlarged. The main
-window has a minimum width of 760 logical pixels; shorter windows scroll
-vertically. Mix master cards wrap onto additional rows when needed.
-Long plugin names are shortened with an ellipsis, with the full name in
-the tooltip. Plugin control and chain windows keep actions below their
-headings, and actions wrap when space is limited.
+The mixer cards follow the window width, up to a limit of 1300 logical
+pixels, so faders and dropdowns keep a usable length on a maximized
+ultrawide instead of stretching across the screen. The window has no
+minimum width of its own: it squeezes down to about 640 pixels, where
+the widest row, the seven input toggles, still fits. A window shorter
+than its content scrolls vertically. Mix master cards wrap onto further
+rows when needed. Long plugin names are shortened with an ellipsis, with
+the full name in the tooltip. Plugin control and chain windows keep
+actions below their headings, and actions wrap when space is limited.
 
 <a name="mic-to-call"></a>
 ### 3.1 Send your microphone to a call or a recording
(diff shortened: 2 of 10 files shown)


CarinaSchoppe:
Thanks a lot for the very detailed follow-up and especially for documenting not only what changed, but why.

I have to admit that reading through all of this is a little humbling 😅. There are definitely a few cases where my original implementation made assumptions that did not hold up, particularly around autostart handling, symlinks, startup repair, the tray behaviour on Linux, and reverting failed toggle changes.

The autostart changes in particular make sense to me. Reusing OpenXlrPaths.WriteAtomic there looked reasonable while I was working through the codebase, but I did not fully account for the fact that ~/.config/autostart is a shared desktop environment directory with very different expectations regarding permissions and symlinks. The revised helper and the more conservative repair behaviour sound substantially safer.

The tray issue is also a good example of something I would probably not have caught without much deeper Avalonia/X11 knowledge. My assumption was basically that tray construction could act as a capability check, which clearly is not true with Avalonia's Linux implementation. Removing the _tray is not null gate therefore makes sense.

Same for the rejected checkbox states. I understood that the underlying preference should remain unchanged on failure, but I missed the binding/UI consequence of raising the property with the same value. The explicit set-and-revert behaviour is a much better solution.

Regarding #81, I am completely fine with the sizing changes being partially reverted. The reasoning and measurements for both removing the 760 px minimum and restoring the 1300 px content cap are convincing. The intent of my changes was mainly to avoid broken layouts, so keeping the parts that actually improve that while reverting the ones that create problems on smaller or ultrawide displays is exactly what I would want.

I also really appreciate the additional tests. The X11 close-path test, layout assertions at multiple widths, and much broader autostart coverage make these behaviours considerably less dependent on assumptions than they were before.

I am still getting familiar with quite a large part of OpenXLR and worked my way through a lot of the architecture while implementing #80–#82, so some of these edge cases were outside what I understood about the project at the time. This follow-up is genuinely useful for me because it shows where those assumptions broke down and how the project expects these cases to be handled.

So no objections from my side to any of the fixes or the two intentional sizing reverts. Thanks for taking the time to review the merged work this thoroughly and for turning the issues into regression tests rather than just patching them.

Wanna merge that straight away or are u fixing a few more stuff?
Cheers M8

emaspa:
Going to merge then work on troubleshooting user issues, probably a new release later today or tomorrow :)

PR #119, opened 2026-09-16
Mixer: make summed monitor gain tests independent of phase

CarinaSchoppe:
The summed Monitor A+B gain regression assumed that two independent combine streams preserve the phase of a 1 kHz sine wave. On CI, different path latencies reduced the summed peak even with correct per-path gains (0.03732 measured versus 0.04372 expected).

Use a constant signal to measure gain independently of phase, discard startup/drain quarters, and require over 95% of settled samples to match within the existing 0.002 tolerance. This also checks sustained audio rather than accepting a single matching peak. Document the test's scope: it does not test frequency response or phase alignment.

Validation: native-enabled Release build with zero warnings/errors. Private PipeWire integration suite on CachyOS: 15 passed, including DSP audio. Ubuntu 24.04 container with older PipeWire/WirePlumber: 14 passed, one opt-in DSP test skipped. Both runs include nine monitor gain/mute/feed scenarios, desktop synchronization and profile startup. No live hardware or installed audio service was changed.


emaspa:
Superseded by #125, which builds on this commit so the authorship is preserved. It keeps the constant signal and adds: the settled window is found from the audio rather than fixed capture quarters, so a long pre-roll on a slow runner cannot put silence in the window; 99% of settled samples must match instead of 95%; and no sample in the whole capture may exceed 1.1 x expected + 0.002, which restores the over-level-anywhere coverage the peak check had. The scope note moved to CONTRIBUTING.md, next to the runner instructions, since native/README.md is about the plugin host. #125 is merged and green.

PR #121, opened 2026-09-16
CI baseline check, do not merge

emaspa:
Temporary. Runs CI on main plus only the OpenDeck needle commit, which touches no audio code, to establish whether MonitorMasterChangesTheRecordedAudioExactlyOnce fails on the current runners independently of the mixer changes in #118. Will be closed and the branch deleted once the answer is in.


emaspa:
Result: failed with the same assertion as #118, peak=0.03732280805706978 against expected 0.0437248. This branch is main plus one commit that touches only the OpenDeck plugin and two docs lines, so MonitorMasterChangesTheRecordedAudioExactlyOnce is red on main on the current runners, independently of any mixer change. Closing; the run log is the record.

PR #127, opened 2026-09-16
Plugins: bound native host protocol buffers

CarinaSchoppe:
Plugin libraries share the native helper stdout. An unterminated line could grow the daemon reader without limit, and arbitrary parameter names could grow its control and meter tables even while the helper kept sending heartbeats.

Read in bounded blocks, discard an oversized line in full, and resume at the next newline. Limit each table to 4096 symbols of at most 255 characters, matching the native host limits. Existing symbols still update at capacity, and draining pending controls frees their slots. The audio path is unchanged; these limits cover daemon protocol buffers, not plugin process memory or CPU use.

This targets development as requested. Its baseline is current main (c6a68f7, 0.1.36), also carried by #126, because development still predates the merged maintenance fixes. After that baseline reaches development, this diff reduces to the three files for this fix. No release version is changed.

Validation:
- Native-enabled Release build: zero warnings and errors.
- Regular .NET suite: 732 passed, 11 opt-in cases skipped.
- Four added regressions cover oversized and fragmented lines, recovery to heartbeats/UI replies, table saturation and draining, maximum-length symbols, empty symbols and non-finite values. The two oversized-output/table-growth cases failed before the fix.
- Isolated PipeWire/DSP runner: all 24 checks passed, including monitor volume synchronization, profile startup, ClipGuard, low cut and a native LSP gate using the changed reader.
- GitHub CI and both CodeQL analyses pass on the pushed commit, including native audio/CLAP/VST3/editor and Xvfb desktop checks. GitHub verifies the commit signature.
- Current NuGet vulnerability data reports no known vulnerable direct or transitive packages in the five projects.

Physical hardware, reboot and interactive third-party plugin editor desktop acceptance were not exercised. The live installation was not replaced or restarted.


emaspa:
Follow-up on the meter table. The cap is right, but nothing checks the names. Controls are filtered against the catalogue when the mixer drains them, and the drain empties the table; meters are never drained. A library that writes to the shared stdout while it loads can fill all 4096 slots with names of its own, after which every real meter the host reports is refused for the life of the process, and the junk is copied into every state snapshot sent to every client. #133 filters meter symbols against the catalogue on arrival, with a test that floods the table and then reports a real meter for the first time, and lists the protocol bounds in docs/api.md beside the other client-visible limits.

One request for future PRs: land a main catch-up as its own merge, and keep the fix to a one-commit PR. #126 and #127 each carried a 46-commit catch-up in front of the real change, which made the diff hard to read without reconstructing it commit by commit.

PR #131, opened 2026-09-17
Plugins: avoid repeated scans through directory links

CarinaSchoppe:
A link back into a plugin search folder caused both CLAP and VST3 discovery to scan the same plugin repeatedly. The regression fixture launched the scanner 41 times for one plugin; branching aliases can multiply the directory traversal further. Recursive CLAP discovery could also lose the whole root when one child was unreadable.

Share a directory walker between the catalogues, using the existing canonical-path resolver to visit each resolved folder once per search root. Preserve the discovered paths, linked roots and nested plugin folders, and keep VST3 bundles terminal. Catch failures around lazy enumeration so readable siblings survive. No arbitrary catalogue size limit is introduced.

This targets development and builds on #130. The additional change is commit 68d5c74; merging #130 first leaves only the directory discovery changes here.

Validation: native-enabled Release build has zero warnings/errors. All 797 regular .NET tests pass, with 11 opt-in cases skipped. Eight new cases cover cycles, linked roots, repeated aliases, an unreadable child, missing roots and terminal VST3 bundles. Both cycle cases failed before the fix with 41 scanner launches instead of one. All 24 separate PipeWire/DSP checks pass. Physical hardware and interactive third-party plugin editor acceptance were not exercised; the live installation is unchanged.


emaspa:
Skipping an unreadable folder instead of losing the whole search root is the right call. The skip was silent, though: `FindBundles` swallowed the exception, so the `directory-error` outcome in the discovery diagnostics could no longer be produced, and a user with a root-owned folder under /usr/lib/clap would send an archive that says the folder was fine. https://github.com/emaspa/openxlr/pull/134 has the walk report each folder it passes over, and the scan records it under its own path. The same PR restores the beating-helper invariant that #130 folded into the poll, by giving that helper a patience a slow reader cannot exhaust.

The shape of #130 to #132 was much easier to review than #126 and #127: one clean catch-up merge, then one commit per PR. Worth keeping.

PR #135, opened 2026-09-17
Keep undeclared meter names out of the native host's table

emaspa:
The same fix as #133, which went to `development` and should have come here.
Adapted for `main`, which has no cap on the meter table yet: here the filter
is what bounds it at all.

Nothing checked the names the helper reported as meters. Controls are
checked against the catalogue when the mixer drains them, and the drain
empties the table; meters are never drained, so a library that writes to the
shared stdout while it loads could fill the table with names of its own,
every one of which was then copied into every state snapshot sent to every
client. The host now keeps the meter symbols the catalogue declares for its
plugin and drops the rest on arrival, with a test that floods it with five
thousand undeclared names and then reports a real meter for the first time.

When `development` is next merged into `main`, this line and the api.md
sentence will conflict with #127 plus #133; take `development`'s side, which
has both the cap and this filter.

Passed!  - Failed:     0, Passed:   793, Skipped:    11, Total:   804, Duration: 52 s - OpenXLR.Tests.dll (net10.0)


emaspa:
Superseded by #136. `main` now carries this fix through #133, in the form that sits on #127's protocol bounds, so this branch's adaptation for a #127-less `main` is no longer needed. Merging it would have meant hand-resolving NativePluginHost.cs, NativeHostProtocolTests.cs and docs/api.md against the same fix arriving from the other side.

PR #138, opened 2026-09-17
Fix: claim the vendor interface before USB control transfers

Michael-Brooks:
## Summary
- The vendor interface (3) on the Wave XLR family is never claimed via `libusb_claim_interface` before control transfers go out, on either the direct in-process path or the default USB-helper-process path.
- Left unclaimed, that traffic contends with the kernel's own USB activity on the shared control endpoint — audible as crackling on the device's audio outputs. Reproduced on a Wave XLR MK.1 across a full-speed USB 1.1 link, with the kernel logging "did not claim interface 3 before use" on every transfer.
- `InProcessUsbTransport.Open()`/`Close()` now claim/release the interface. No detach is needed since no kernel driver owns interface 3 on any model in the family, so this doesn't disturb the audio-class interfaces the kernel driver still owns.

## Test plan
- [x] Verified live on a Wave XLR MK.1: crackling on device audio outputs gone with the fix applied, present without it.
- [x] Confirmed via kernel log: "did not claim interface 3 before use" no longer logged once claimed.


emaspa:
Thanks, taking this. One correction to the explanation before it lands, so the comment in the code says what the kernel does.

usbfs already claims the interface itself: for a class request addressed to an interface, `check_ctrlrecip` calls `checkintf`, which logs "did not claim interface 3 before use" once per open file and then binds the interface to usbfs (`claimintf`). Every later transfer on that handle finds the bit set and passes silently. For vendor-type requests (the Pro, the MK.2 and the Dock MK.2 dialects) it returns before any interface check. So an explicit claim is the same kernel state the first transfer would have produced, and it changes nothing about what goes over the wire. If the crackling on your MK.1 went away in an A/B, something else moved between the two runs. The 100 ms poll on a full-speed link is where I would look next if it comes back.

The claim is still worth doing: it removes the one warning line on the MK.1 and the Dock, and it states which interface the daemon owns. I pushed one commit on top of yours that rewrites the doc comment to that effect and makes a failed claim say why on stderr instead of reading as a missing device. Squash-merging once CI is green.

PR #139, opened 2026-09-17
Fix: newly created mixer sinks could come up muted

Michael-Brooks:
## Summary
- `Mixer.Build()` never asserted a mute state on the mix-bus, channel, and post sinks it creates, relying on PipeWire/WirePlumber's default for a freshly loaded node. A same-named node's prior mute state can carry over across recreation, so a daemon restart could silently mute the whole graph regardless of `mixer.json`.
- Same root cause in a third, previously untouched code path: `CreateFilterChain` (used by both the mic low-cut/ClipGuard DSP chain and mix insert chains) never asserted its capture sink's mute state either — a muted filter-chain sink silences everything downstream even with healthy hardware capture feeding it, which is what made a mic appear dead in the UI despite real signal reaching the input.

## Fix
- Explicitly call `SetSinkMuted` after each sink is created: mix busses get their configured `Muted` state; channels, post sinks, and filter-chain sinks are always unmuted (none has a whole-sink mute concept of its own).

## Test plan
- [x] Verified the mic no longer comes up silenced after a daemon restart with the low-cut/ClipGuard filter chain active.


emaspa:
Thanks for tracking this down. The premise is right: WirePlumber's stream-properties state restores mute and volume by node.name for these sinks, and a same-named sink can come back muted after a restart.

This branch is based on 0.1.12, and main has handled it a different way since 0.1.27. MixerService runs a one-second sweep (`Mixer.SyncOwnSinkLevels`) that unmutes and resets to full volume every `OpenXLR_` sink except the monitor mixes, which covers the channel, post and filter-chain sinks (`OpenXLR_lc_*_in`, `OpenXLR_ins_*_in`). Build reapplies the monitor mix mute from the model through `ReapplyMixLocked`. So on 0.1.38 the sinks this PR touches are unmuted within a second of coming up, and the only window left is the one before the first sweep.

Two things here would also work against main as it stands now: mute on a stream or virtual-mic mix lives on the combine legs, not the sink, so asserting the configured mute on the sink at build would be undone by the next sweep; and an unguarded pactl failure inside Build would fail the whole graph, where the sweep catches it.

Closing as superseded. If a mic still comes up silent on 0.1.38 with the filter chain active, please open an issue with the daemon log and I'll look at closing that first-second window at creation.

PR #140, opened 2026-09-17
Fix silent mic on the original Wave XLR (MK.1) and a startup-defaults race

Michael-Brooks:
## Summary
Root-causes and fixes a silent-mic report on the original Wave XLR (MK.1, 0fd9:007d), plus a startup race that could point the fix's own defended value at the wrong default.

**Silent mic on MK.1** — the card's pro-audio profile groups its capture and playback ALSA nodes into one PipeWire node group (confirmed via `pw-dump`). If a playback stream opens on the card before its capture stream is established, the mic goes silent for that capture stream's entire life: bit-exact zero, structurally normal, no errors — while a plain `arecord` against the same hardware in isolation always captures real audio. This is the same full-duplex ordering bug already fixed for the XLR Dock (`50-xlr-dock-capture-hold.conf`).

Verified live, repeatedly:
- solo capture: always real audio
- playback opens first, capture joins after: silent, reliably
- capture opens first, playback joins after: stays real audio, including racing playback within 0.3s of daemon start
- full cold restart of pipewire/pipewire-pulse/wireplumber/daemon, with the fix in place: real audio every time

Fix: a WirePlumber rule (`packaging/52-openxlr-mk1-capture-hold.conf`) that keeps the MK.1's capture node from ever suspending, so it's always the side already running when anything else opens playback on the card — the same approach already proven for the Dock. Wired into the Debian and RPM packaging alongside the existing Dock/Pro rules.

**Startup-defaults race** — found while chasing the report above. `MixerService.StartAsync` captured "what the default sink/source were before we touched anything" by calling `pactl` at the top of its own `StartAsync`, to defend that value once WirePlumber's delayed auto-switch tries to reassign the newly-created nodes. But `DeviceManager` is a separate hosted service starting concurrently, and connecting the device can flip the card to the pro-audio profile, which makes WirePlumber recreate the card's ALSA nodes — and sometimes reassign the system defaults to them — before `MixerService` gets to take its snapshot. When that race goes the wrong way, the daemon spends the rest of the session defending an already-wrong default: looks fixed after a user corrects it by hand mid-session, quietly reverts on the next boot, which is the timing where the race is most likely to lose.

Fix: move the snapshot to the top of `Program.cs`, before any hosted service exists, and hand it to `MixerService` via DI.

## Test plan
- [x] Verified end-to-end against a Wave XLR MK.1 across repeated cold restarts, with and without the low-cut filter chain active.


emaspa:
Thanks, both halves check out and I'm taking them.

The race is real on main today: DeviceManager is a BackgroundService registered before MixerService, its first sweep parks the card in pro-audio (`TryParkCardProfile`), and MixerService still took the snapshot inside its own StartAsync. The rule is MK.1-specific as written: the MK.2, the Pro and the Dock MK.2 all enumerate under "Elgato" rather than "Elgato Systems" (lsusb in #2 and the Pro on my desk), so the "Elgato_Systems_" prefix leaves them alone.

The branch was based on 0.1.13 and conflicted with the MixerService and Program.cs of 0.1.38, so I rebased it onto main, kept both commits under your authorship, and added one of mine that installs the rule from the Nix package and mentions it in the source-install steps, the manual and the hardware support page (the deb and rpm already had it from your commit). Squash-merging once CI is green.

PR #142, opened 2026-09-18
Development: add flexible routing and desktop controls with daemon hardening

CarinaSchoppe:
OpenXLR can now send any mix to selected outputs, combine independent mix-to-output routes with a level per route, and use additional PipeWire capture sources as input channels. Desktop and OpenDeck actions route the focused application, adjust output volume or mute, and select the enforced default output. The daemon maintains an incremental PipeWire registry instead of starting a full graph dump on every sweep.

This integrates the six feature PRs #143, #145, #146, #147, #148 and #144, plus capture recovery in #151, with the existing development hardening. The current main branch's Deck dial responsiveness and baseline bridge compiler-flag changes are preserved. Settings, live commands, state and user-facing behavior are documented in the existing API, manual and layout documentation.

The integration covers these failure paths:

- Output routes retain independent gains across profile recall, deletion and reconnect. Internal matrix nodes stay out of output choices. Capture inputs require both sink connections; a partially successful connection is removed and retried rather than being reported as healthy.
- Portal shortcut sessions ignore delayed replies and signals from replaced or disabled sessions. Rapid presses use one ordered consumer with at most 16 waiting actions; overload is reported and stale queued actions are discarded. Focus lookup reports unsupported desktops or missing application identity rather than routing an arbitrary application.
- Registry updates publish atomically with bounded framing, depth, object and text budgets. The observer reconnects after failure, discards stale state and waits for observed sink writes. Normal routing and watchdog reconciliation remains in place.
- USB helper exchanges have deadlines and failed protocol streams are discarded. Shutdown stops pending settings writes. Plugin catalogue parsing handles malformed nested fields, invalid ids, non-finite controls, reversed ranges and escaped-message size limits while retaining valid entries.
- App assignment limits are enforced under the mixer lock. Stream identity reuse, failed moves, forgotten assignments and case-insensitive aliases are handled without retaining an incorrect live route. UI state refreshes preserve controls without echoing assignments, update layout order, and coalesce obsolete meter frames.

Performance changes reuse catalogue indexes and immutable graph snapshots, avoid redundant parsing on the sweep, and bound pending UI work. The registry regression serves 10,000 unchanged reads of a synthetic 2 MB graph with no managed allocation on the reading thread. This does not establish total daemon CPU, memory or audio latency on every system.

Validation:

- Locked restore and native-enabled Release build with warnings treated as errors: zero warnings and errors.
- The combined application tree including the main branch Deck change passes 999 general managed tests and seven Node tests. The subsequent main change only touches the bridge packaging helper; three direct checks confirm its baseline compiler flags while preserving ordinary hardening flags. The Windows bridge was not rebuilt in this pass. Twenty environment-dependent cases are skipped in the general run; the separate audio and desktop runners are reported below.
- The integrated feature tree passes 32 private PipeWire/profile/DSP tests with no skips, including actual recorded audio, registry-helper termination and reconnect, routing recall/deletion, output keys, and partial capture-link recovery.
- Tray, layout, tooltip and skin tests pass under Xvfb. Native audio, CLAP, VST3 and editor tests, manifest, shell, version, locked-restore, OpenAPI and RPM checks have passed. GitHub CI and CodeQL are checked on the current PR head before merge.
- New deterministic regressions failed before the fixes: six late-session reply cases, two rapid-key ordering/overflow/disable cases, and the partial capture-link case. Private D-Bus and WebSocket fixtures exercise the session lifecycle and actual daemon command acknowledgements.

Physical USB hotplug and multiple attached interfaces, a complete login/reboot cycle, real desktop shortcut permission/activation, live compositor focus acceptance, and manual third-party plugin-editor interaction were not validated in this pass. Audio tests use a private PipeWire server and desktop window tests use Xvfb; they do not replace those checks. No installed daemon or live audio configuration was changed. Versions and release outputs are unchanged.


emaspa:
Follow-up on main after this merge: PR #153.

Integration-wide changes:
- The startup default snapshot waits up to 10 s for pactl instead of 3 s, so a socket-activated pipewire-pulse on a cold login cannot leave the earlier default undefended.
- The catalogue footprint is measured (the entry's wire bytes under the daemon's JSON options, cached per plugin) instead of estimated. The estimate ran about twice over the truth and trimmed real catalogues from the picker early.
- The manual's new sections carry fixed anchors (`capture-inputs`, `desktop-keys`, `output-keys`, `output-matrix`) and the two new windows link to them. The registry and scanner limits prose moved out of the manual; the limits stay in the code.
- The desktop keys tests skip when `dbus-daemon` is not installed instead of failing.

Per-feature changes are noted on the feature PRs.

PR #143, opened 2026-09-18
Mixer: listen to any mix on selected outputs

CarinaSchoppe:
Outputs can now listen to Stream, Chat, Aux and custom virtual microphones, including their insert processing, through the existing feed command and window picker. OpenDeck cycles the live mix list after the existing A, B and A+B choices. Distinct mix ids can also be summed through the API, and the window retains that selection.

Deleting a mix removes it from output feeds as part of the saved layout change. Remaining mixes stay selected; an empty feed returns to the first monitor mix. Failed saves restore both the mix and its feeds. Rebuilding a mix's insert chain also reconnects its listening outputs. The Pro's direct mic path is restricted to monitor-only feeds so listening to a recording mix cannot silence its microphone recording; summed monitor feeds update each individual send. Aux faders remain available when listening to Aux on hardware without a USB Aux port, with no unavailable port switch.

Validation:

- 925 regular managed tests passed; 13 opt-in cases are covered by the private PipeWire and desktop runners.
- 26 private PipeWire/profile/DSP cases passed. Recorded audio covers Stream, Chat, Aux and a Monitor+Chat sum. Live graph checks cover recall, failed-save rollback, removal from a summed feed and deletion of a custom mix used as the sole feed.
- Tray, layout, tooltip and skin tests passed in separate Xvfb processes; the layout check also ran at 2560x1440.
- Five Node tests passed, including feed cycling through new mixes and recovery from a removed choice. Native audio, CLAP, VST3, scanner and editor checks passed, as did manifest, shell, version, lock, OpenAPI and RPM checks.
- The private audio runner accepts OPENXLR_TEST_FILTER for focused regression runs without duplicating the dotnet filter argument.

An initial full audio run missed the existing app-route test's three-second move deadline; two subsequent full runs passed. Physical Pro headphone routing and manual third-party plugin-editor acceptance have not been exercised. Desktop checks used Xvfb and audio checks used a private PipeWire server. No installed application, live audio setup, release version or release output is changed.


emaspa:
Checked on main after the merge; nothing changed here. Two things to know for later work: every jack output now gets an explicit `monitorFeeds` entry, so "absent means primary" no longer holds for jacks, and routing intent lives in `_monitorFeeds`, `_outputRouteLevels` and `_routeGains` together, reconciled by hand in each caller.

PR #144, opened 2026-09-18
Daemon: discover PipeWire objects through registry events

CarinaSchoppe:
The daemon now keeps one `pw-dump --monitor --no-colors` subscription instead of launching and reparsing full dumps on its routing sweep. Registry batches update immutable objects by id. Queries share the cached snapshot until a change arrives. Standalone adapter callers keep a short-lived one-shot cache, now scoped to their adapter.

The subscription waits for the initial core-sync batch, drops its registry on disconnect, and reconnects with bounded backoff. JSON framing handles split UTF-8 and escaped strings, rejects incomplete or malformed batches, and enforces depth, batch, registry-text and object-count budgets. Updates publish atomically. The streaming helper drains bounded stderr and is stopped on cancellation or reader failure. Sink writes wait for their observed state so the next sweep cannot restore a pre-write volume. The existing routing, device and watchdog reconciliation timer remains.

Validation:

- Locked restore and native-enabled Release build passed with zero warnings and errors.
- 999 regular managed tests passed on the combined feature tree; 20 opt-in cases were skipped by that general run. Private audio and Xvfb coverage is reported separately below; live desktop acceptance remains outstanding. The subscription regression run passed all 14 focused tests after the retry-backoff adjustment.
- 32 private PipeWire/profile/DSP cases passed, with monitor controls and captured audio using the subscription. A dedicated integration test kills the owned observer process and verifies fresh device discovery after reconnect.
- A 2 MB synthetic registry serves 10,000 unchanged snapshot reads with zero managed allocations on the reading thread. This measures the registry read path, not total daemon CPU or RSS.
- Stream-assignment tests run the normal reconciliation sweep and check both the tracked channel and the actual PipeWire destination, including delayed publication after a move.
- Regression coverage includes both tombstone forms, id reuse, snapshot lifetime, EOF, cancellation, stderr flooding, split JSON frames, depth/byte limits and cumulative object/text limits.
- Four desktop Xvfb classes, native audio/CLAP/VST3/editor checks, Node tests and the manifest, shell, version, locked-restore, OpenAPI and RPM checks passed.

Integration retains the output matrix, arbitrary capture sources, focused-app routing and output keys from the other feature PRs. Their audio tests now run with the registry subscription enabled, including partial capture-link recovery from #151. Merge #151 first.

No installed daemon or live desktop audio was changed. Physical hotplug and manual third-party editor interaction were not exercised. No package dependencies, release versions, commands or API fields change.


emaspa:
Changed on main in PR #153:
- While the subscription is reconnecting, the adapter reads the graph with a one-shot `pw-dump` instead of throwing from `Read()`. Before, a subscription that stayed down (a registry over the 8 MiB budget, or any exception outside the catch list) left every sweep throwing at the first graph read: no stream routing, no default enforcement, no route repair, and no restart, since nothing reported the graph as gone.
- Own-sink writes no longer throw when the registry does not echo within a second, and they skip the wait while the subscription is down. pactl has already applied the write at that point; a late echo only means the next sweep reads the same value. The throw aborted whole sweep passes from `SyncMonitorVolumesLocked` and the route branch of `EnsureOwnSinkLevelsLocked`.
- A regression test runs the adapter with a subscription that never becomes ready and checks that devices and own sink levels still come from the one-shot dump.

PR #145, opened 2026-09-18
Mixer: add per-route levels and an output matrix

CarinaSchoppe:
Outputs can now receive several mixes at independently saved levels. The Output matrix window edits each route from off to 100%, while the existing output master retains its separate boost range. Internal gain sinks stay hidden and are created only when attenuation is needed.

This PR builds on #143. Its temporary base is `codex/listen-to-any-mix` to keep this diff limited to the matrix. Retarget to development after that dependency merges.

Shared Pro return jacks are represented as one bus. Explicitly silent outputs survive recall and unrelated mix deletion. Missing gain stages and unplugged outputs are reconciled with their saved levels; failed layout saves preserve the previous routing intent.

Validation:
- Native-enabled Release build: zero warnings and errors. Full .NET suite: 936 passed, 14 opt-in tests skipped.
- Private PipeWire server with DSP enabled: 27 passed, including recorded audio for independent route gains, silence, output masters, recall, gain-node recreation and hotplug.
- Xvfb layout, tray, tooltip and skin checks passed. Matrix checked at widths 600, 1000 and 1800, with scrolling and slider interaction; rendered preview inspected.
- Native audio, CLAP, VST3 and editor regression checks passed. Node tests and syntax, manifest, shell, version, locked restore, OpenAPI and RPM checks passed.

No physical multi-output hardware or manual third-party plugin editor acceptance was performed. Shared-jack assertions exercise logical routing, not the physical Pro. The installed daemon and desktop audio were not changed.

Output grouping and per-route gains share one canonical bus-key calculation in the mixer. The cleanup passes 16 focused matrix/feed tests and all 27 private PipeWire/DSP audio/profile tests, with a warning-free native-enabled Release build.


emaspa:
Changed on main in PR #153:
- `SetOutputRoute`'s fast path records the level before writing the gain node. With the write after it, a registry echo that arrived late threw before the level was stored and the next sweep wrote the old level back, reverting the slider.
- The sweep compares route levels at percent resolution. pactl takes whole percents and the registry reports them rounded, so a level like 0.455 was rewritten every second forever.
- The output matrix window has a Manual button and the manual section a fixed anchor.

PR #146, opened 2026-09-18
Mixer: add independent PipeWire capture inputs

CarinaSchoppe:
The layout editor can add an independent input from an external PipeWire capture source and stereo pair. Several microphones, capture cards or Wave interfaces can feed their own channels. Inputs start muted, reuse the existing sends and profile controls, and reconnect only to their saved source. Missing sources or stereo pairs stay silent.

Capture channels cannot receive application assignments. Renaming preserves their graph; failed saves preserve the previous layout. A full capture-only saved layout reserves an application destination even when a discarded capture has the same id and name as the synthesized System channel. Regression cases cover collisions at and beyond the 32-channel limit.

The existing active-device selection still owns hardware controls and built-in input DSP. This change does not add concurrent hardware control or additional per-input plugin hosting. API, layout and user documentation describe these boundaries.

Validation:
- General suite: 928 passed, 13 opt-in skips.
- Private PipeWire/DSP suite: 26 passed, covering simultaneous sources, independent gains, initial mute, missing pairs, removal/reconnection, settings recall, rename, failed saves and deletion without removing the source device.
- Xvfb layout, tray, tooltip and skin checks passed; the capture dialog was exercised at 360 pixels.
- Native audio/CLAP/VST3/editor, Node, syntax, manifest, shell, version, locked-restore, OpenAPI and RPM checks passed.

Physical capture cards, simultaneous physical Wave interfaces and manual third-party plugin editor acceptance were not tested. Audio tests used isolated PipeWire sockets; the installed daemon and live audio were not changed.


emaspa:
Checked on main after the merge; the code is unchanged. The manual section got its fixed anchor (`capture-inputs`) in PR #153.

PR #147, opened 2026-09-18
Desktop: route focused apps from portal and OpenDeck keys

CarinaSchoppe:
A focused application's audio can be routed to an application channel from an OpenDeck Toggle key or a PC global shortcut. The opt-in Desktop keys window uses the GlobalShortcuts portal and stays active in the tray. A short-lived KWin query supplies the process id; the daemon resolves an exact or unambiguous child-process PipeWire identity and reuses the existing remembered assignment.

The portal does not expose focused-window identity, so focused routing currently requires KDE Plasma. Unsupported desktops, absent clients and ambiguous processes produce explicit errors. No window titles are read. OpenDeck keys receive success/error feedback. The UI must keep running, and the daemon needs GLib's gdbus command.

Method timeouts and cancellation close the private desktop connection to release outstanding replies. Deck-only sessions also report connection loss. Apply reconnects after the desktop service recovers. User and API documentation explain the behavior and prerequisites.

This PR builds on #146's application/capture distinction. Retarget to development after that dependency merges.

Validation:
- General suite and private D-Bus tests passed, covering portal creation/binding/activation, invalid keys/sessions, cancellation, closed sessions, save failure, an unresponsive peer, timeout and Deck-only disconnect.
- Private PipeWire/DSP suite previously passed all 27 tests. After expanding the routing regression, all 26 non-DSP audio/profile tests passed again. Routing is tested both before the first mixer sweep and after discovery, with the daemon's normal sweep running while waiting for the real destination.
- The read-only KWin query passed in the live KDE session and removed its temporary script. It did not route live audio, change focus or register real global shortcuts.
- Xvfb layout/tray/tooltip/skin, native audio/CLAP/VST3/editor, Node and repository format/packaging checks passed.

The actual desktop permission dialog/key binding and physical Stream Deck hardware were not exercised. Portal activation and Deck key messages were tested through isolated D-Bus and WebSocket connections. Xvfb does not replace manual plugin editor acceptance.

Shortcut replies and queued portal callbacks are scoped to their current session. Six delayed-reply regressions cover success and failure after disabling, closing and replacing a session, then verify activation after reconnecting. All six failed before the correction and pass afterwards. The combined six-feature general suite passes 997 tests (19 opt-in skips).


emaspa:
Changed on main in PR #153:
- The KWin scripting calls ride a connection of their own, opened per query. `DesktopBus.AwaitReply` disposes its connection on a timed-out reply, and with KWin on the shared connection one slow `loadScript` (2 s, easy under a full-screen game) disposed the portal session and the `org.openxlr.Desktop` name too, so every shortcut stayed dead until the user reapplied. KWin still reports back on the shared connection.
- `gdbus` is declared where the daemon is packaged: `glib2` in the RPM spec, `libglib2.0-bin` in debian/control, `glib` on the Nix wrapper's PATH. The AUR PKGBUILD needs `glib2` when it is next touched.
- Apply while a configuration is still in the portal dialog is queued behind it instead of being dropped silently; the status says so.
- The six tests that start `dbus-daemon` skip where it is absent instead of failing.
- docs/api.md states that `assignApp` and `assignStream` take application channels only.

PR #148, opened 2026-09-18
Mixer: control PipeWire outputs from desktop and Deck keys

CarinaSchoppe:
Output keys can adjust a named PipeWire sink or the current system default, toggle mute, and select the enforced default output. PC keys use the shared GlobalShortcuts portal session; OpenDeck uses the same acknowledged daemon commands. Five-point volume steps support 0 to 150%, monitor mix masters and linked monitor outputs without stale client-side slider values.

The selector retains missing devices, rejects internal unity-gain sinks and ambiguous Pulse aliases, and preserves capture-default policy and mixer feeds. The first volume key after selecting monitor outputs establishes the synchronization baseline before writing. Monitor mute keys read the current server state and reuse the existing mix setter; external sink mute uses an atomic server toggle. Shared desktop timeout/disconnect cleanup and routing startup coverage from #147 are included. API, HTTP and user documentation accompany the commands.

This branch is stacked on #147. Retarget to development after that dependency lands. Merge order: #143 before #145; #146 before #147 before this PR. #144 is independent. When combining #144 with the other features, retain both the private route-node filter in ListDevices and WaitForSinkLevel in PipeWireAdapter.

Validation:
- Locked restore and native-enabled Release builds: zero warnings/errors.
- Output-key branch: 955 general tests passed, 16 opt-in skips; 28 private PipeWire/DSP audio/profile tests passed.
- Combined six-feature worktree: 991 general tests passed, 19 opt-in skips; 31 private-server audio/profile tests passed with registry subscriptions enabled for the new paths.
- Audio checks cover default switching, 0/150% bounds, repeated mute, independent monitor masters, rejected targets and first-press follower synchronization.
- Seven Node tests passed, including plugin message dispatch and unavailable devices. Private D-Bus tests exercise shortcut registration and activation.
- All four Xvfb desktop classes, native audio/CLAP/VST3/editor and repository format/packaging checks passed. Narrow selectors and rendered labels were checked.

Physical Wave interfaces, Stream Deck hardware and a real desktop permission/key-binding session were not exercised. Xvfb editor checks do not replace manual plugin controls, move, resize and reopen acceptance. No installed audio service or configuration was replaced. Combined validation used a separate unpublished worktree.

Shortcut replies and queued portal callbacks are scoped to their current session. Six delayed-reply regressions cover success and failure after disabling, closing and replacing a session, then verify activation after reconnecting. All six failed before the correction and pass afterwards. The combined six-feature general suite passes 997 tests (19 opt-in skips).

Rapid PC shortcut presses use one ordered consumer with a maximum of 16 waiting actions. Full queues report overload; disabled or replaced sessions discard waiting actions. Private-portal tests reproduce the old dropped-press behavior, verify output-switch/volume order after a failed command and prove that disabling discards pending work. The merged branch passes 985 general tests (18 opt-in skips); all 13 desktop-key tests pass on the final queue implementation.


emaspa:
Changed on main in PR #153, so the mute key and the Deck dials agree with each other:
- `toggleOutputMute` on one of the selected monitor outputs toggles the mixes feeding that output, the same thing a dial press on the monitor does. Before, the key muted the physical sink while the dial muted the monitor mix master, and neither showed the other's state. A monitor mix sink still toggles its mix, and any other output still toggles the sink at the audio server.
- New `setOutputDeviceVolume` (optional `device`, `value` 0 to 1.5) sets an output's level outright, so a dial can drive it with the plugin's echo logic.
- Sink entries in `devices` carry `volume` and `muted`, and the daemon pushes state when any sink's level or mute changes. The Deck output mute key lights red while its output is muted, and a new "Outputs" dial group turns an output's volume up to 150% and mutes it on press, ring included. The dial's existing monitor press now mutes the mixes feeding the first selected monitor output instead of always Monitor A.
- Manual and API docs updated for all of the above.

PR #151, opened 2026-09-18
Mixer: retry incomplete capture connections

CarinaSchoppe:
If PipeWire accepts only one of a capture channel's stereo links, the mixer previously retained that partial route as connected. Reconciliation then checked only the retained side and never attempted the missing side again.

Require both sink inputs before retaining a capture route. Remove partial links and let the existing sweep retry the complete connection. Mono inputs still feed both sides. The layout documentation describes the behavior.

Validation:
- A private PipeWire fault-injection test rejects the right-side link once. It failed against the old code and now verifies disconnected state, removal of the partial route and recovery of both links.
- The general suite passed 971 tests with 18 opt-in skips before rebasing onto current development; the warning-free native-enabled Release build and targeted private-server regression passed again after rebasing.
- No physical hardware or installed audio service was changed.


emaspa:
Checked on main after the merge; nothing changed here.

PR #156, opened 2026-09-18
Mixer: customize channel and mix presentation

CarinaSchoppe:
Customize channel and mix icons, colours and display order, hide channels without removing their audio routing, and use a compact view with one selected channel. Stream Deck keys use the same appearance metadata.

Display order and routing order remain separate. The layout editor's **Use displayed order for routing** button uses the existing `setLayoutOrder` command. Display arrows do not change the fallback for unassigned apps. `primaryMonitorMix` keeps default output feeds independent of display order in the window, terminal, Omarchy bar and Deck.

Profiles now capture the mixer appearance map. A missing map preserves current presentation, an empty map clears it, and entries for deleted targets are dropped. Malformed data rejects the profile before device settings change. Profiles saved from the window also carry its skin, compact view, selected channel and collapsed sections. Section order is transported for the dependent window-order PR #191. Startup, update, tray and security preferences remain local.

A successful recall has a unique presentation revision, persisted with the local window choices. Repeated state frames, reconnects and window restarts therefore preserve manual edits made after recall. Older clients can overwrite profile audio without discarding saved window choices. Unavailable skins fall back to the default. Failed compact, collapse and skin preference writes retain the visible choice and report an error.

Validation after integrating main `d10794e` on `d07be0b`:

- Locked restore and Release build: zero warnings and errors.
- 1,278 general tests passed; 20 environment-gated tests skipped.
- Separate X11 skin-window suite passed, including profile recall of an invalid installed skin.
- Private PipeWire audio suite: 31 passed, one optional ClipGuard test skipped.
- Deck tests, manifest validation, version/locked-restore checks, OpenAPI, RPM specification and Omarchy QML/skin checks passed.
- Regression coverage includes profile round trips, missing and malformed fields, failed recalls, deleted targets, recall revision persistence, repeated recalls, preference write failures and preservation of unrelated local settings. The missing mixer appearance recall was reproduced before correction. Scene unit tests use private fake helpers and do not require the desktop audio packages.

Physical hardware and Stream Deck acceptance remain open. The conflict in `docs/mixer-layout.md` is resolved by retaining the upstream rollback/retry behavior and the presentation/profile documentation. The stale statement excluding presentation from profiles is corrected. Other feature PRs still need integration against the exact main they eventually target.

Fresh GitHub CI and both CodeQL analyses passed on `d07be0b`. GitHub verifies the signed integration commit. No PR was merged.


emaspa (changes requested):
Reviewed against development at 7adb041. This matches the per-mix customization item on the roadmap and is the first of the seven to take.

Findings:

- `src/OpenXLR.UI/ViewModels.cs:566` The window's move buttons now send `setDisplayOrder` only and never `setLayoutOrder`. Config order still decides the default destination for unassigned streams (`MixerLayout.cs:118`), so moving a channel to the top in the editor no longer changes which channel new applications land in, and nothing else in the window sets that any more. Either keep sending `setLayoutOrder` for editable items and let the display order follow it, or keep the two apart and give the editor a control for the routing order. The hint at `MixerSetupWindow.axaml:54` should then say how the routing order is set.
- `docs/manual.md:1681` and `docs/mixer-layout.md:90` carry the same four paragraphs of window walkthrough. The layout file doc keeps the `appearance` map paragraph; the walkthrough belongs in the manual only.
- `src/OpenXLR.UI/LayoutAppearanceViewModel.cs:39` and `:50` load `ui.json` in two more field initialisers, so the main view model reads it three times at construction. One load assigned to the three fields.

Checked and fine: rollback on a failed save, orphan entries dropped after delete, validation on both sides, the plugin re-validating icon and colour before they reach the SVG, hidden refused on mixes, no new appearance token needed, command registered, dispatched and documented, test class in the xdg-config collection.

Landing: every one of the seven open branches conflicts with every other one (docs/api.md, docs/features.md and docs/manual.md at least), so this one merges clean only because it goes first. The others need a rebase after it lands.

emaspa:

Why:

- #156: `docs/features.md`, `docs/http-api.md`, `docs/mixer-layout.md` (the output routes section became output feeds).

PR #157, opened 2026-09-18
Plugins: consolidate search paths and rescans in the plugin manager

CarinaSchoppe:
Plugin Manager keeps search directories and rescanning in one place. Additional paths are saved atomically and remain removable when a drive is offline; removing a path keeps its plugin files.

lilv retains its compiled defaults when LV2_PATH is unset, including Nix store paths. Custom LV2 bundles are discovered separately in that case; live hosts require an explicit LV2_PATH to use those directories. Resolved aliases and overlapping known paths are rejected, broad system roots are refused, and recursive discovery has a reported entry limit. Duplicate additions do not rescan. Disconnect clears a stale picker selection.

An explicit LV2_PATH is applied to catalogue discovery and child hosts even without additional directories, including after the managed environment changes. Without an explicit value, lilv retains its compiled defaults. This also avoids reading the additions file twice to construct the override.

A registered directory that later becomes a symbolic link loop cannot hide healthy custom roots. Unresolvable entries produce a warning and block edits until repaired, preserving the complete saved configuration. Healthy entries remain available regardless of entry order.

Validation: Release build with zero warnings/errors; 1,271 general tests passed, 19 environment-gated skips; 31 targeted search-path and LV2 bundle tests passed, including real lilv discovery. Tests cover compiled defaults, explicit overrides, aliases, missing directories, failed saves, concurrent edits, scan bounds, broken saved paths and recovery. The saved-directory and explicit-path regression cases failed before correction. The final combined integration passes 1,271 general tests (37 opt-in skips) and all 61 isolated PipeWire audio tests with the fixture and system LV2 paths supplied. The existing X11 layout check remains applicable to the unchanged UI. The CI change only adds search-path tests to the lilv step.

No physical device or running desktop acceptance is claimed. This PR targets current main.

Updated onto main aea487e (0.1.46), preserving the Dock MK.2 change and its documentation. Locked restore and the Release build pass with zero warnings/errors. Fresh CI and CodeQL pass on the rebased head; GitHub verifies all feature commit signatures. No PRs were merged.


emaspa (changes requested):

Findings:

- `src/OpenXLR.Core/Mixing/PluginSearchPaths.cs:75` and `:101` When `LV2_PATH` is unset and the user adds one LV2 folder, the daemon replaces lilv's compiled-in default with a hand-written FHS list. Where lilv's default is not that list (a flake-only NixOS install, where lilv is built with store paths; the packaged module sets the variable so it is fine) every system LV2 plugin disappears from the catalogue and from new loads the moment a custom folder is added. `Snapshot()` also lists this hand-written set as the effective paths even when no override applies and lilv is using its own. The override has to add to lilv's default rather than replace it. If the daemon cannot learn that default, leave the variable unset when it was unset, scan the extra folders in the catalogue only, and say in the manual that a filter-chain load from a custom folder needs `LV2_PATH`.
- `PluginSearchPaths.cs:19` normalises with `GetFullPath` only. A symlinked folder and its target both register, and a custom folder under a default one is scanned twice, so a CLAP or VST3 bundle can appear twice in the picker. Resolve links and refuse a folder that lies under an existing search directory. The duplicate test only covers `/.` and a trailing slash.
- `PluginSearchPaths.cs:20` accepts any absolute directory except `/`. Adding `/usr` or `/proc` as a VST3 path recurses the whole tree at every start and every rescan with no guard.
- `src/OpenXLR.Daemon/WebSocketHub.cs:315` adding a path that is already registered returns Ok and still runs the full catalogue refresh.
- `src/OpenXLR.UI/InsertsViewModel.cs:160` clearing the plugin choices and selection on disconnect is a behaviour change for a user with the picker open when the daemon restarts. Fine if intended; say so.
- `src/OpenXLR.Tests/PluginSearchPathTests.cs:14` sets `CLAP_PATH` and `VST3_PATH` process-wide for the class's lifetime. The xdg-config collection does not serialise this against other classes that read the catalogue search path.
- `docs/manual.md:1709` to `:1716` describe internals (replies from an old daemon connection, controls refreshing after a rescan). That belongs in the pull request, not the manual.
- `.github/workflows/ci.yml:53` adds the new test class to the lilv step. Acceptable as it is test scope only; keep the CI edit to that line.

Checked and fine: settings written atomically with private mode, commands registered, dispatched and documented, no new tokens, the test class in the xdg-config collection, offline removal, corrupt file preserved, failed write, concurrent additions and multiarch paths each have a test.

Landing: this conflicts with every other open branch, so it needs a rebase once #156 is in.

emaspa:

Why:

- #157: `docs/features.md`.

PR #158, opened 2026-09-18
Mixer: report plugin latency and add optional mix compensation

CarinaSchoppe:
Report algorithmic latency per insert and optionally align parallel mix outputs. Compensation defaults off, is bounded to two seconds and waits for valid reports. Only a mix needing positive correction receives a delay stage; layouts without positive reported latency gain no extra stage.

Stage creation and live mix creation check file-descriptor headroom. Repair retains healthy plugin instances, prunes obsolete reports and preserves both errors when rewiring and persistence rollback fail. Saved and live compensation use the same host-selection path. The unrelated allocation probe is extracted into #197, leaving one shared implementation.

Validation on `92718f4`:

- 1272 general managed tests passed; 25 environment-gated tests were skipped in that run.
- All Deck tests passed.

Fresh GitHub CI and CodeQL are being checked on this published head. Hardware acceptance is not claimed by the general test suite. Physical monitoring acceptance and a real VST3 plugin changing latency without reactivation remain open. The existing changes-requested review still needs the maintainer's decision; this update does not dismiss it. No PR was merged.


emaspa (changes requested):

Findings:

- `src/OpenXLR.Core/Mixing/MixLatency.cs:53` and `src/OpenXLR.Core/Mixing/LiveMixerLayout.cs:67` The delay nodes sit outside the pipewire-pulse open-file accounting. Enabling compensation creates one filter-chain module and two nodes per mix with no `EnsurePulseHeadroomLocked` call, and creating a mix while it is on reserves two nodes although four appear. On a layout near the limit the option can push pipewire-pulse over its limit, which the headroom check exists to prevent. With no inserts anywhere every mix still carries a delay node and an extra hop; the stage should exist only for a mix whose chain reports latency.
- `docs/manual.md:1683` has curly quotes, `:1686` an arrow, and `:1708` to `:1715` describe implementation (the identity index, the transient delay-control retry). `:1693` says a missing native helper or unsupported host feature is shown in Options, but `MixLatency.Plan` only ever exposes the generic waiting message, the delay-node errors or a `set-param` exception.
- `MixLatency.cs:39` if `WireInputFeedsLocked` throws and the rollback `PersistLocked` also throws, the second exception replaces the first and the in-memory setting is restored while the saved file may say the opposite.
- `MixLatency.cs:100` and `Mixer.cs:436` the "toggle compensation to retry" message is wrong: once the restart window lapses the sweep retries on its own.
- `MixLatency.cs:141` stale latency reports are pruned only when the total insert count changes. Remove one insert and add another between sweeps and the old key stays.
- `native/vst3.cpp:548` dropping the restart on `kLatencyChanged` matches the vendored header, but a plugin that resizes its buffers only on `setActive` would report a latency it is not yet applying. Needs a check with a real plugin; note the acceptance gap in the description.
- `src/OpenXLR.Tests/PipeWireGraphTests.cs` the allocation probe rewrite is also in #161. Keep it in one of the two.
- `.github/workflows/ci.yml:47` switches the native build to the native-enabled managed build and builds the fixture there. The fixture rule belongs with the tests; the build-path switch is a separate change. #159 and #160 add the same line at the same place, so the three conflict.
- Commits: seven, with a fixup ("Tests: ignore the compiled latency fixture") and one that bundles two changes. Squash to two or three, with bodies that say why.

Checked and fine: latency reporting stays on the helper's main thread with no audio-thread allocation, the command is registered, dispatched and documented, only existing tokens are used, helpers use `ExecutableScript.Write`, the integration test is in the xdg-config collection, and the allocation assertion is not weakened.

Landing: conflicts with every other open branch, and with #159 and #160 in Mixer.cs, PipeWireAdapter.cs, MixerService.cs and WebSocketHub.cs. Rebase after the earlier ones land.

emaspa:

Why:

- #158: `CONTRIBUTING.md` (the check list now matches the CI workflow line for line; add your line to main's list), `docs/features.md`, `src/OpenXLR.Core/Mixing/PipeWireAdapter.cs`.

PR #159, opened 2026-09-19
Mixer: audition microphone processing with Sound Check

CarinaSchoppe:
Sound Check records up to ten seconds of dry microphone audio in memory, loops it through the live software chain, and restores the physical input on stop. Only one microphone session is allowed; samples never enter profiles or saved settings.

Missing-host failures explain this feature. Errors are scoped to the session's microphone and explicit stop clears them. Device changes, graph rebuilds, missing audio and expiry restore the physical feed with a reason; failed startup and failed restoration retain both errors. UI disconnect and stale replies cannot restore a superseded session.

Validation on `586991d`:

- 1264 general managed tests passed; 21 environment-gated tests were skipped in that run.
- All Deck tests passed.

Fresh GitHub CI and both CodeQL language analyses passed on this published head. GitHub verifies the commit signatures. Hardware acceptance is not claimed by the general test suite. Physical microphone and live desktop acceptance remain open. The existing changes-requested review still needs the maintainer's decision; this update does not dismiss it. No PR was merged.


emaspa (changes requested):
Reviewed against development at 7adb041. The audio path holds up: no allocation, lock or file access on the audio thread, buffer bounds hold, the NaN command falls to live, device change, helper death, broken link and expiry all restore the physical feed, and the state stays out of settings exports.

Findings:

- `src/OpenXLR.Core/Mixing/SoundCheck.cs:31` on a build without the native helper, Record fails through the `NativePluginHost` constructor with "switch this insert back to the filter chain". The window shows it verbatim and the second sentence is wrong for this feature. A dedicated message, or a `HostInstalled` check before starting.
- `SoundCheck.cs:20` and `Mixer.cs:222` the session error is only cleared when a new recording starts. After "stopped because the input device changed" or the ten-minute expiry it stays in every state broadcast until someone presses Record again, and `SoundCheckViewModel.cs:33` copies it regardless of channel, so XLR 1's error shows in the XLR 2 window. Clear it on stop and scope it to the session's channel.
- `Mixer.cs:2224` a rebuild ends the session with no reason set, so the window just flips to "Live microphone" unlike every other end path.
- `SoundCheck.cs:51` if `WireInputFeedsLocked` throws during Record, the rollback calls it again and a second throw replaces the original exception.
- `Mixer.cs:421` the first sweep reports a change for every user (null against idle) and triggers one extra scan and state push at start. Initialise the last state to idle.
- Tests: the managed ten-minute path and a UI disconnect during a loop have no test (only the native clock test), and the "other microphone" refusal has none.
- `.github/workflows/ci.yml:48` adds the native-enabled managed build and `test-sound-check`. The native test hook belongs here; the build line is added by #158 and #160 too at the same place and the three conflict.

Landing: conflicts with every other open branch, and with #158 and #160 in the mixer core. Rebase after the earlier ones land.

emaspa:

Why:

- #159: `docs/features.md`, `src/OpenXLR.Core/Mixing/PipeWireAdapter.cs`.

PR #160, opened 2026-09-19
Mixer: support inserts on every input channel

CarinaSchoppe:
Allow stereo-compatible insert chains on application, external-capture and Aux channels before their existing mix sends, using the same commands and editor workflow as microphone and mix inserts.

Reworked on current main: empty software/capture chains retain the original combine sink and add no bus or route. The first insert creates a split path only for that channel; removing its final insert restores direct sends. The public name and assignments remain stable, but those two transitions recreate the sink and can interrupt that channel briefly. Bypass and edits within an existing chain retain its public sink. Unchanged profile/settings recall keeps software plugin instances running. A failed processing or direct-fallback route is reported on its own channel and does not abort other channel rewires. XLR strips no longer duplicate their existing Inserts button.

Healthy link pairs come from an index of the existing registry snapshot, with no pw-link process for an already connected pair. The index is reused while the snapshot is unchanged. Existing repair still handles absent pairs and disappearing ports.

Independent work is split into #186 (native CI and ClipGuard), #187 (LV2 host fallback with accurate live host state) and #188 (editor movement/reopening checks). This branch includes the shared #186 and #187 parents so its CI also exercises Ubuntu without a working filter-chain LV2 loader; feature fixture setup stays here.

Feature validation before this main integration: native-enabled Release build, zero warnings/errors; 1,254 general tests passed and 25 gated checks skipped on the final dependency stack. Eighteen isolated PipeWire checks passed, including software/capture audio on both sides, preserved plugin IDs across recall, removal/rename/save rollback, monitor controls, and 100 healthy link checks with zero link-helper launches. The final eight targeted channel/link checks passed, as did the prior X11 layout and skin checks. New fault injection checks both processing-link failure with a healthy direct fallback and failure of both routes. Another channel still starts its effects and retains its instance during recovery. The healthy fallback previously prevented retrying the failed plugin; that regression failed before the fix and passes now.

The maintainer's roadmap decision for application/Aux inserts remains outstanding. Hardware and a real desktop session have not been accepted here. No removed per-output gain API is restored.

Local audio-test instructions now include the gain fixture and LV2_PATH required to discover it, plus installed plugin paths for the DSP cases. CI already supplies the fixture path. The runner allows five minutes for the expanded combined suite; individual helper deadlines remain unchanged. With all open features integrated, the full private PipeWire run passes all 61 cases in 2m43s, close to the previous three-minute suite limit.

CI passed before the main update. Its first run had one timeout in the existing fake native-editor protocol test; the unchanged rerun passed every stage. The complete local suite also passed with DOTNET_PROCESSOR_COUNT=2, as did ten fresh-process repeats of that specific test. The sporadic timeout remains unexplained; production editor deadlines were not relaxed.

Integrated current main `d10794e` in signed commit `9ecb585`. The `docs/mixer-layout.md` conflict retains both channel-insert behavior and upstream deletion rollback/retry behavior. Locked restore and native-enabled Release build pass with zero warnings/errors. The current general suite passes 1,258 tests with 26 gated checks skipped; all four channel-insert integration cases pass separately on private PipeWire sockets with the documented gain fixture. Fresh GitHub CI and both CodeQL analyses pass on this head, GitHub verifies the integration signature and the PR is conflict-free. No PRs were merged.


emaspa (changes requested):
Reviewed against development at 7adb041. Not closing this, but it needs rework, and a roadmap decision before that: the roadmap defines plugin inserts as chains on each XLR input and every mix, and the mixer block says the routing model settles in users' hands before further plugin expansion. Application and Aux inserts are not listed, and this changes the graph shape for every install.

Findings:

- `src/OpenXLR.Core/Mixing/ChannelInserts.cs:15` and `Mixer.cs:163` every software and capture channel gets the null sink plus the hidden bus whether or not it carries effects. On the default layout that is six extra sinks, six stereo links, an extra graph hop on every application channel and about sixty more pipewire-pulse file descriptors on the base count, for a user who never adds an effect. The bus has to exist only for a channel with effects, and a layout without them has to produce the same node set as 0.1.41, with a test that asserts it.
- `ChannelInserts.cs:67` with `PipeWireAdapter.cs:1173` the sweep now calls `EnsureLinks` for every software channel every second, and `EnsureLinks` spawns `pw-link` per pair and treats "File exists" as healthy. Default layout, no effects: twelve process spawns a second that all fail with EEXIST, on the path the roadmap already names as the daemon's CPU problem. Link health has to come from the graph dump the sweep already parses, not from a spawn.
- `Mixer.cs:1157` and `:1253` profile recall runs `WireChannelChainLocked` for every software channel, which starts with `RemoveChannelChainLocked`, so recalling a profile interrupts audio on every application channel and restarts running plugins whose definition did not change.
- `ChannelInserts.cs:63` runs outside the try. A three-second port timeout in `LinkStereoNodes` at build time now fails the whole graph, where a software channel needed no link before, and a `setInserts` on channel A can throw before B and C in the same command are rewired.
- `PipeWireAdapter.cs:692`, `:785` and `:828` the native fallback is remembered per plugin URI for the daemon run and consulted for every later chain, so the XLR 1 chain then also runs through the hosted path while the insert's `NativeHost` stays false (`Mixer.cs:772`). The state and api.md do not show it and the editor cannot be opened. This fallback is useful on its own for PipeWire builds without the LV2 loader (Ubuntu's 1.0.5) and would be taken as a separate pull request, with the state saying which host runs the insert.
- `src/OpenXLR.UI/MainWindow.axaml:628` the new Inserts button is in the strip template XLR 1, XLR 2 and Aux also use, so the XLR strips show two Inserts buttons.
- `.github/workflows/ci.yml` adds two packages, a second full build and a ClipGuard step, and `native/tests/lsp-editor.py` and `native/README.md` extend an unrelated desktop check. Split these out; the ClipGuard step would be taken as its own pull request.
- `docs/manual.md` has work-log sentences (the three-second wait, partial links removed before retrying, remembered for the daemon run). The manual says what the user sees.

Landing: conflicts with every other open branch, and with #158 and #159 in the mixer core.

emaspa:

Why:

- #160: `docs/mixer-layout.md` (the output routes section became output feeds), `src/OpenXLR.Core/Mixing/PipeWireAdapter.cs`.

PR #161, opened 2026-09-19
Plugins: copy chains, save presets and compare A/B settings

CarinaSchoppe:
Copy effects and whole chains between compatible channels, rename instances, save whole-chain parameter presets and compare A/B snapshots. Copies have independent data and fresh instance IDs. Per-plugin preset libraries, file import/export and plugin-private binary state remain outside this change.

Local or daemon-side edits invalidate listening labels, including edits racing a pending acknowledgement. Queued parameter changes are sent before replacing a chain. Integration with main retains its shared SliderSync.FlushPending and ReleaseTouchGuards helpers and one timer path, alongside the chain-specific flush. No second slider implementation is added.

Validation on `9d89095`:

- 1268 general managed tests passed; 20 environment-gated tests were skipped in that run.
- All Deck tests passed.

Fresh GitHub CI and both CodeQL language analyses passed on this published head. GitHub verifies the commit signatures. Hardware acceptance is not claimed by the general test suite. No new physical hardware or live Wayland/Deck acceptance is claimed. The existing changes-requested review still needs the maintainer's decision; this update does not dismiss it. No PR was merged.


emaspa (changes requested):

Findings:

- Scope: this is whole-chain presets only. The roadmap item reads "per-plugin and whole-chain, with export and import; copy a chain between channels; A/B comparison". There is no per-plugin preset and no export or import, so the item stays open after this merges. Either add them here or say in the description what remains.
- `src/OpenXLR.Tests/PipeWireGraphTests.cs:62` the allocation probe rewrite is also in #158. Keep it in one of the two; the second to merge conflicts on it.
- `docs/manual.md:1700` says Sound Check can keep looping while these settings change. Sound Check is #159; if this merges first the manual describes something that does not exist.
- `src/OpenXLR.UI/EffectWorkflow.cs:40` appending past sixteen inserts fails with "Unsupported effect chain version, width or size". The count check needs its own message.
- `EffectWorkflow.cs:57` "Hearing A" is set after a recall and nothing clears it when the user then moves a parameter or edits the chain.
- Interaction with #162: A/B recall and preset apply go through `setInserts` with bypass in the snapshot, and #162's `SetInserts` replaces the client's bypass with the pre-hold value for any insert under a hold. With both merged a held key makes an A/B switch lose its bypass changes. One of the two has to handle it; I have noted it there too.

Checked and fine: the preset file goes through `WriteAtomic` with private mode and a damaged file is never overwritten, clipboard and snapshots are deep copies with fresh ids on paste and load, width is left to the daemon, stale replies are dropped by the connection epoch, rename touches only the label and still broadcasts and saves, control windows follow the instance and the same id on two channels gets independent windows, the new window uses Ox.* resources only, commands registered, dispatched and documented.

Landing: conflicts with every other open branch. Rebase after the earlier ones land.

emaspa:

Why:

- #161: `docs/features.md`, `src/OpenXLR.Daemon/CommandValidation.cs`.

PR #162, opened 2026-09-19
Plugins: support held effect keys with automatic restoration

CarinaSchoppe:
Momentary effect keys activate an insert or chain while held and restore its earlier state on release. Overlapping holds share restoration, leases cover lost releases, and temporary bypass changes do not enter profiles or saved settings.

Explicit chain replacement cancels holds and respects the client's requested bypass values. Renew/end do not wait for plugin installation. Profile and settings recall compare against actual running bypass before clearing holds, including partial and older profiles. Socket shutdown relies on lease expiry rather than an unflushed end message.

Validation on `cf3700e`:

- 1261 general managed tests passed; 24 environment-gated tests were skipped in that run.
- All Deck tests passed.

Fresh GitHub CI and both CodeQL language analyses passed on this published head. GitHub verifies the commit signatures. Hardware acceptance is not claimed by the general test suite. The maintainer's roadmap decision and physical Stream Deck acceptance remain open. The existing changes-requested review still needs the maintainer's decision; this update does not dismiss it. No PR was merged.


emaspa (changes requested):
Reviewed against development at 7adb041. Not closing this, but held keys are not on the roadmap, and plugin work follows the mixer block there. It needs a roadmap entry before it is taken, and it lands after the plugin items already listed.

Findings:

- `src/OpenXLR.Core/Mixing/Mixer.cs:771` `SetInserts` replaces the client's bypass with the saved pre-hold value for any insert kept by id and plugin, and `Begin` records a saved entry for every insert in the hold, including ones that were already active. Whole-chain key held, insert X active, a client sends `setInserts` with X bypassed: X stays active, the change is silently dropped and the hold is cancelled. Today only HTTP clients hit it since the window uses `setInsertBypass`, but #161's A/B recall and preset apply go through `setInserts`. Override only the inserts the hold actually flipped, or apply the client's value after cancelling.
- `src/OpenXLR.Daemon/WebSocketHub.cs:314` renewals and ends take `_installGate`, the lock an install, sync or delete holds for its whole duration. A key held during an install longer than five seconds sees its renewals queue while the sweep expires the hold, so the effect drops mid-press. Only `begin` needs the gate.
- The docs call it "Held effects" in api.md, "Momentary effect keys" in the manual and "Held effect keys" in features.md. One name.
- `plugin/com.emaspa.openxlr.sdPlugin/plugin.mjs:148` the clear before `process.exit` cannot flush the socket, so the end frames are lost and the lease covers it anyway. Drop the call or say that in the comment.
- Tests: the `keyDown`, `keyUp` and `willDisappear` wiring in plugin.mjs and the inspector checkbox are untested, and nothing covers the `setInserts` case above.
- Two commits, the second docs-only. Squash.

Checked and fine: overlapping holds restore only when no other hold covers the insert, a hold on a removed insert or deleted chain is dropped cleanly, manual bypass and profile recall cancel holds, expiry writes no settings, renewal traffic is well inside the command budget, repeated key down is idempotent on both sides, a late error for a superseded hold cannot release the new press, non-momentary keys are untouched, the command is registered, dispatched and documented.

Landing: conflicts with every other open branch, and with #161 in CommandValidation.cs, MixerService.cs and WebSocketHub.cs.

emaspa:

Why:

- #162: `docs/features.md`, `src/OpenXLR.Daemon/CommandValidation.cs`.

PR #166, opened 2026-09-19
Nix: compile the bridge companion for the baseline x86-64

emaspa:
The Debian, RPM and AUR companions are built by build.py, which strips
the packager's target flags and compiles the host for the baseline
x86-64, because GCC spills AVX registers around the host's calls into
Windows code and the reload faults. The Nix companion builds the bridge
from nixpkgs' derivation and never saw that flag, so a platform with
gcc.arch set got the faulting build. The derivation now appends the
same target to its compiler flags.

Stacked on #163 for the one README sentence. Checked by evaluating the flake: the bridge derivation's NIX_CFLAGS_COMPILE carries the target. A full build of the companion was started locally; its result is reported in a comment when it ends.


emaspa:
The companion built from this branch with nix build .#openxlr-yabridge: the receipt reads 5.1.1.54-2 and the bridge library carries no AVX instructions (no ymm or zmm registers and no vzeroupper in its disassembly).

emaspa:
Closed by GitHub when its base branch was merged and deleted; continued as https://github.com/emaspa/openxlr/pull/168 against main.

PR #167, opened 2026-09-19
Mixer: take out the per-route output levels

emaspa:
An output's feed already names any mix or a sum of mixes, and every
mix in a feed reaches the output as a direct port link that PipeWire
sums. The per-route level on top of that, shipped in 0.1.40, was a
second mixing layer over the sends: it cost a hidden null sink per
route below unity, its own window, command, state fields, saved routes
and scene fields, and a second answer to what an output hears. A blend
at other levels is what a mix is for.

The feed picker, the sums, the shared-jack identity and the incomplete
route repair stay. The setOutputRoute command is gone, outputRoutes is
ignored in the settings file and in profiles, and an empty saved feed
now reads as the first monitor mix, since nothing writes one any more.
The docs describe the feeds as they are and say what became of the
routes.

Stacked on #163 for the docs it rewrites. Verified with the full managed suite, the private PipeWire audio runner (the summed-feed capture now checks unity links on two outputs) and the window layout test under Xvfb. The desk daemon was not rebuilt in this pass. A user who stored a silent output through the matrix in 0.1.40 or 0.1.41 hears the first monitor mix there after upgrading and unticks the output instead.


emaspa:
Closed by GitHub when its base branch was merged and deleted; continued as https://github.com/emaspa/openxlr/pull/169 against main.

PR #184, opened 2026-09-20
Devices: detect the XLR Dock MK.2 USB control bank

CarinaSchoppe:
The XLR Dock MK.2 can enumerate normally and play audio while every vendor settings read fails with `Pipe error`. The backend assumed bank `0x0103`; an owner reports a `0fd9:00c7` unit returning the expected 38-, 2- and 6-byte blocks at `0x0203` instead.

Probe the three known blocks on connect, preferring `0x0103` and trying `0x0203` only after a stall or unexpected length. Detection performs no writes. Keep the selected bank for all controls until disconnect, detect again on reconnect, and close the handle if detection fails. Timeout, disconnect and hung-helper errors do not trigger another bank. Options reports the selected bank through the existing connection note.

Validation: Release solution build with zero warnings and errors; 1,242 general tests passed with 19 environment-gated skips before the final test-only addition; all 15 targeted bank tests pass on the final commit, covering both banks, incomplete blocks, retries, transport failures, reconnect and write failure. The supplied September 20 Omarchy archive confirms the repeated settings-read failure and an available PipeWire device. Its empty block dump does not verify the alternate bank; successful `0x0203` reads and the manual workaround are owner-reported.

Hardware acceptance remains open: I have not run automatic detection on either physical Dock MK.2 variant. The original `0x0103` hardware verification is retained in the hardware-support document, with the new report and acceptance gap distinguished.


emaspa:
Merged and released in [0.1.46](https://github.com/emaspa/openxlr/releases/tag/v0.1.46). Thanks for the detection work.

Two changes went in on top before the merge. A dock that answers on neither bank now stays open on `0x0103` instead of having its handle closed: the probe demanded the exact block lengths while `Read` right below it accepts any block long enough for the offsets it indexes, so a short-but-usable unit that works today would have stopped connecting, and a closed handle also puts `DumpBlocks` out of reach, which is what a third revision needs before anyone can support it. The connection note is written only when the dock departs from the usual bank, since `DeviceManager` logs it at warning level and Options shows it for as long as the device is connected.

Two things recorded in `hardware-support.md` while checking the premise. Both revisions report `bcdDevice 2.10` with otherwise identical descriptors, so nothing short of the transfers tells them apart. Vendor interface 3 has an alternate setting 1 with two interrupt endpoints, so "interface 3 has no endpoints" described setting 0, the one the control transfers use.

Hardware acceptance is still open on both variants.

